PR #2, opened 2023-06-13
first template

sofianhnaide:
This is the initial commit for the anyscale templates repo.

A repo made specifically for templates that are anyscale-specific, similar to the ray-specific repo under: https://github.com/ray-project/ray/tree/master/doc/source/templates.


templates/anyscale-ray-101/README.md
emmyscode, line 1: These images aren't rendering for me.

emmyscode (commented):
Reposting from Slack:

What's the difference between creating a template in the ray-project vs. anyscale ? Is the default to contribute to OSS ray-project unless there's some proprietary features to show off?
Or is it the case that most templates should live in both?

templates/anyscale-ray-101/README.md
sofianhnaide: here you go, if you have a different suggestion that using `base64` I am open for it.

Screen Shot 2023-06-13 at 12 48 27 PM

emmyscode: Got it!

sofianhnaide:
> Reposting from Slack:
> 
> What's the difference between creating a template in the ray-project vs. anyscale ? Is the default to contribute to OSS ray-project unless there's some proprietary features to show off? Or is it the case that most templates should live in both?

I would say anything that is generic should go to ray (ray tune, serve, train, data, etc..). If you are doing something specific to anyscale, e.g. Anyscale jobs, or something that does not make sense having in Ray, e.g. Aviary, then this would go to the anyscale templates.
This repo is still public, so this is not for proprietary content. If we have content like that, we can add another repo for that. It does not make sense to have the template in both. I would always start by asking does it fit in Ray, if the answer is no, then contribute to the Anyscale template repo

gvspraveen:
Dumb question. For ray specific templates we have doc/source/templates/templates.yaml which specifies cluster env. 

```
batch-inference-ray-data:
...
  cluster_env:
    build_id: anyscaleray-ml250-py39-gpu
  compute_config:
    GCP: doc/source/templates/configs/compute/gpu/gce.yaml
    AWS: doc/source/templates/configs/compute/gpu/aws.yaml
```

do we have templates/templates.yaml here in this repo?

sofianhnaide:
> Dumb question. For ray specific templates we have doc/source/templates/templates.yaml which specifies cluster env.
> 
> ```
> batch-inference-ray-data:
> ...
>   cluster_env:
>     build_id: anyscaleray-ml250-py39-gpu
>   compute_config:
>     GCP: doc/source/templates/configs/compute/gpu/gce.yaml
>     AWS: doc/source/templates/configs/compute/gpu/aws.yaml
> ```
> 
> do we have templates/templates.yaml here in this repo?
The `doc/source/templates/templates.yaml ` in the Ray repo is not actually used, @justinvyu I think we should probably remove it since the source of truth will always be in the product.

justinvyu:
Yeah, it's not used, but the Ray CI does check that all of the paths are valid, which isn't done on the product. I'm ok to just remove it on Ray.

README.md
```diff
@@ -0,0 +1,32 @@
+# Anyscale Starter Templates
+
+These templates are a set of minimal examples & tutorial for customers to run on the anyscale platform.
```
kamil-kaczmarek, line 3:
```suggestion
These templates are a set of minimal examples & tutorials for customers to run on the anyscale platform.
```

README.md
```diff
@@ -0,0 +1,32 @@
...
+
+    ```text
+    templates/my-awesome-template
+        <name-of-your-template>/
+            README.md
+            <name-of-your-template>.ipynb/py
+    ```
+
+    Your template does not need to be a Jupyter notebook. It can also be presented as a
+    Python script.
```
kamil-kaczmarek, line 24: Can this be mix of both?

For example: https://github.com/ray-project/ray-educational-materials/tree/main/NLP_workloads/Text_generation

README.md
```diff
@@ -0,0 +1,32 @@
...
+
+    All templates MUST have a `README.md` file.
+
+2. Add your compute configuration under `configs/<your-template-name>` (for both AWS and GCE).
```
kamil-kaczmarek, line 28: Can I also add "cluster environment"?

This is necessary for most of our content.

kamil-kaczmarek:
Can we add templates specifically to some Anyscale organizations?

Say *orgA* has some templates, *orgB* has some other templates? I can think of these templates being delivery channel for training events!

README.md
sofianhnaide: yes it can, if you plan to add a cell at the end of the notebook to deploy as a job or as a service, you have to convert the notebook into python or include a python version since anyscale job & services don't support notebooks.

README.md
sofianhnaide: You can add a BYOD, meaning build your own docker image, publish it and users can use it.

The reason we are not supporting cluster envs now is for the build time, we want users to be able to click no a template and start it right away. It's a little bit more work for now on the template authoring to build the image and publish it, but that's to get the best experience on the user side. If you need help publishing docker images, I am happy to help.

sofianhnaide:
> Can we add templates specifically to some Anyscale organizations?
> 
> Say _orgA_ has some templates, _orgB_ has some other templates? I can think of these templates being delivery channel for training events!

Yes you can, please check the instructions in: https://www.notion.so/anyscale-hq/Workspace-Templates-df285c5560504163bfbc824a9a75c034?pvs=4#6069b9d41042450ca48295802154be8e

kamil-kaczmarek (approved):

PR #10, opened 2023-08-01
Update import in trt-resnet.py

akshay-anyscale:
Importing tensorrt outside the actor fails when using serve deploy since it requires GPU

### DIFF
--- a/templates/image-classification-service/trt-resnet.py
+++ b/templates/image-classification-service/trt-resnet.py
@@ -7,7 +7,6 @@ from typing import Dict
 
 import torch
 import time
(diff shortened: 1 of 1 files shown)


templates/image-classification-service/trt-resnet.py
```diff
@@ -22,6 +21,7 @@
 @serve.ingress(app)
 class Classifier:
     def __init__(self):
+        import torch_tensorrt
```
edoakes, line 24: suggest to leave a comment here explaining sole delayed import

edoakes (approved):

Issue #14, opened 2023-11-07
Add multi-lora to workspace template


GeneDer:

GeneDer:
Also need this to use the new image https://github.com/anyscale/product/pull/24881

PR #15, opened 2023-11-07
Add an example to run lora models

GeneDer:
Add an example to run lora models. Also update code to use the new openai api.

Closes: https://github.com/anyscale/templates/issues/14


GeneDer:
Tested with the new image on workspace for the old query and the new lora base model, static lora, and dynamic lora all working as expected :) 

image

templates/endpoints/README.md
```diff
@@ -151,6 +153,60 @@ Endpoints allows you to easily add new models by adding a single configuration f
 To learn more about how to customize or add new models, 
 see the [Model Registry](models/README.md).
 
+# Serving LoRA Models
+
+`serve_lora.yaml` and `query_lora.py` are provided for you in this template.
+- If you just want to deploy static LoRA models, you can list them in `multiplex_lora_adapters` config in the `serve_lora.yaml` file.
+- If you want to dynamically loading LoRA models to a running service, you can use `dynamic_lora_loading_path` config in the `serve_lora.yaml` file.
```
akshay-anyscale, line 160:
```suggestion
- If you want to dynamically load LoRA models to a running service, you can use `dynamic_lora_loading_path` config in the `serve_lora.yaml` file.
```

templates/endpoints/README.md
```diff
@@ -151,6 +153,60 @@ Endpoints allows you to easily add new models by adding a single configuration f
...
+- Both can be set at the same time and complement to each other.
```
akshay-anyscale, line 161:
```suggestion
- Both can be set at the same time and complement to each other. The lora adapters specified in "multiplex_lora_adapters" will be always loaded in each serve replica whereas the ones from "dynamic_lora_loading_path" will be loaded as needed.
```
Is this accurate?

templates/endpoints/README.md
```diff
@@ -151,6 +153,60 @@ Endpoints allows you to easily add new models by adding a single configuration f
 To learn more about how to customize or add new models, 
 see the [Model Registry](models/README.md).
 
+# Serving LoRA Models
+
+`serve_lora.yaml` and `query_lora.py` are provided for you in this template.
```
akshay-anyscale, line 158: We should add something to the order of - We support serving multiple LoRA adapters with a common base model in the same request batch. In addition, we use serve multiplexing to reduce the number of swaps for LoRA adapters.

templates/endpoints/README.md
```diff
@@ -151,6 +153,60 @@ Endpoints allows you to easily add new models by adding a single configuration f
...
+#         "total_tokens": 395
+#     }
+# }
+```
+
+A few tips for serving LoRA models:
+1. LoRA base models should be passed in the serve config file `serve_lora.yaml` in the `multiplex_models` config.
+1. `dynamic_lora_loading_path` in `serve_lora.yaml` can be loaded from any AWS S3 or Google Cloud Storage bucket where the workspace has access to. You can use an existing bucket where you have the loRA models or can use `$ANYSCALE_ARTIFACT_STORAGE` already provided by Anyscale Workspace.
+1. LoRA checkpoints can be added to the `dynamic_lora_loading_path` dynamically before and after the Serve is already started.
+1. LoRA checkpoints can to be stored in the `{base_path}/{base_model_id}:{suffix}:{id}` format (e.g. `s3://my-bucket/my-lora-checkouts/meta-llama/Llama-2-7b-chat-hf:lora-model:1234`), where the `base_path` is defined in `dynamic_lora_loading_path` config and `base_model_id` should match with one of the models in `multiplex_models`.
```
akshay-anyscale, line 206:
```suggestion
1. LoRA checkpoints have to be stored in the `{base_path}/{base_model_id}:{suffix}:{id}` format (e.g. `s3://my-bucket/my-lora-checkouts/meta-llama/Llama-2-7b-chat-hf:lora-model:1234`), where the `base_path` is defined in `dynamic_lora_loading_path` config and `base_model_id` should match with one of the models in `multiplex_models`.
```

templates/endpoints/README.md
```diff
@@ -151,6 +153,60 @@ Endpoints allows you to easily add new models by adding a single configuration f
...
+#     "usage": {
+#         "prompt_tokens": 285,
+#         "completion_tokens": 110,
+#         "total_tokens": 395
+#     }
+# }
+```
+
+A few tips for serving LoRA models:
+1. LoRA base models should be passed in the serve config file `serve_lora.yaml` in the `multiplex_models` config.
```
akshay-anyscale, line 203: Everything is numbered 1 in this list

templates/endpoints/models/meta-llama--Llama-2-7b-chat-hf_a10.yaml
```diff
@@ -42,3 +43,5 @@ scaling_config:
   placement_strategy: "STRICT_PACK"
   resources_per_worker:
     accelerator_type_a10: 0.01
+multiplex_config:
```
akshay-anyscale, line 46: Add comment explaining this and how to set it

templates/endpoints/multiplex_lora_adapters/lora_config.yaml
```diff
@@ -0,0 +1,13 @@
+base_model_id: meta-llama/Llama-2-7b-chat-hf  # <-- this is the base model for this loRA
```
akshay-anyscale, line 1: lora_config.yaml -> static_lora_adapter_config.yaml

templates/endpoints/README.md
GeneDer: Actually I think both are only loaded when the user made the call to the model 😅

templates/endpoints/README.md
GeneDer: Yep, this is a trick for writing numbered list in markdown. It will automatically increment the number when it's rendered so if we insert any thing in between, we don't need to change the number for all :)  
image

templates/endpoints/README.md
akshay-anyscale: ok , then we should clarify that the adapters are only loaded when the model is called

templates/endpoints/models/meta-llama--Llama-2-7b-chat-hf_a10.yaml
```diff
@@ -43,5 +43,5 @@ scaling_config:
   placement_strategy: "STRICT_PACK"
   resources_per_worker:
     accelerator_type_a10: 0.01
-multiplex_config:
+multiplex_config:  # <-- This is required to be used as a multiplexed model
```
akshay-anyscale, line 46: It's unclear to an unfamiliar user what multiplexed is .. Something like - This is required to be able to serve multiple LoRA adapters in the same replica.

templates/endpoints/README.md
akshay-anyscale: This makes the "dynamic" vs "static" wording confusing.

templates/endpoints/README.md
```diff
@@ -41,6 +41,7 @@ RayLLM leverages Ray Serve, which has native support for autoscaling and multi-n
 - [Deploying as a Production Service](#deploying-on-anyscale-services)
 - [Using the OpenAI SDK](#using-the-openai-sdk) 
 - [Model Registry](#model-registry)
+- [Serving LoRA Models](#servomg-lora-models)
```
akshay-anyscale, line 44:
```suggestion
- [Serving LoRA Models](#serving-lora-models)
```

templates/endpoints/README.md
```diff
@@ -151,6 +153,56 @@ Endpoints allows you to easily add new models by adding a single configuration f
...
+#     "usage": {
+#         "prompt_tokens": 285,
+#         "completion_tokens": 110,
+#         "total_tokens": 395
+#     }
+# }
+```
+
+A few tips for serving LoRA models:
+1. LoRA base models should be passed in the serve config file `serve_lora.yaml` in the `multiplex_models` config.
```
akshay-anyscale, line 199: Numbering still seems off

templates/endpoints/README.md
```diff
@@ -151,6 +153,56 @@ Endpoints allows you to easily add new models by adding a single configuration f
...
+#     ],
+#     "usage": {
+#         "prompt_tokens": 285,
+#         "completion_tokens": 110,
+#         "total_tokens": 395
+#     }
+# }
+```
+
+A few tips for serving LoRA models:
```
akshay-anyscale, line 198:
```suggestion
These are the requirements for serving LoRA models:
```

akshay-anyscale (approved):
small comments, but looks good overall

templates/endpoints/README.md
GeneDer: nice catch!!

templates/endpoints/README.md
GeneDer: This will render correctly in markdown 

something like 
```
1. foo
1. bar
1. baz
```
will turn into 
1. foo
1. bar
1. baz 
😄

PR #16, opened 2023-11-09
Stable diffusion and llama2 inferentia template

akshay-anyscale:


templates/inferentia-llama/llama-serve.py
```diff
@@ -0,0 +1,33 @@
...
+    def generate(self, prompt):
+        inputs = self.tokenizer(prompt, return_tensors="pt")
+        outputs = self.model.generate(**inputs,
+                                     max_new_tokens=128,
+                                     do_sample=True,
+                                     temperature=0.9,
+                                     top_k=50,
+                                     top_p=0.9)
+        return self.tokenizer.batch_decode(outputs, skip_special_tokens=True)
+
```
gvspraveen, line 30: i think we should consider a streaming example at some point for llm use case. Probably can be done as follow up

gvspraveen (approved):

PR #33, opened 2024-01-03
Add embedding model

akshay-anyscale:


templates/endpoints/models/embedding_models/BAAI--bge-large-en-v1.5.yaml
```diff
@@ -0,0 +1,33 @@
+deployment_config:
```
gvspraveen, line 1: do we need to update readme about embedding models?

akshay-anyscale: updated

gvspraveen (approved):

PR #49, opened 2024-01-29
[serve] More small cleanup to stable diffusion template

edoakes:


scottsun94 (approved):

templates/serve-stable-diffusion/README.ipynb
```diff
@@ -35,11 +35,12 @@
    "metadata": {},
    "source": [
     "### Step 2: Run the model locally\n",
-    "- Open a new terminal (ctrl+shift+` in VSCode.)\n",
-    "- Run the command below to deploy your model at http://localhost:8000.  \n",
+    "- Run the command below in a VSCode terminal.\n",
```
scottsun94, line 38: Curious about why we remove the keyboard shortcut?

edoakes: oh I removed it because it's not actually needed, there's already a terminal window open that you can use by default

scottsun94: I think we are not gonna open the terminal by default after we switch to ipynb instead of readme.. We may need to add that back

edoakes: oh got it, will re-test and update once that happens

PR #58, opened 2024-02-01
Add intro services template

ericl:
Adds an intro services template

@scottsun94 , after this merges could you also add it to the templates file with a similar config as the Intro to Workspaces template?

@edoakes , let me know your feedback. I'm not a services expert though so ideally someone on the team could take a pass afterwards to improve the notebook.


scottsun94:
LGTM. Do we land on "serve deploy" or "serve publish"?

templates/intro-services/README.ipynb
```diff
@@ -0,0 +1,287 @@
...
+    "print(send_request(\"Theodore\"))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Monitoring production services\n",
+    "\n",
+    "Along with the monitoring tools that come with Workspaces, in Services you also get a number of built-in metrics out of the box in the `Metrics` tab. This tab includes aggregated metrics across all rollouts for the service (possibly from multiple Ray clusters).\n",
```
akshay-anyscale, line 184: We should add a note here about using the serve logger as well (and in the code example).

edoakes:
> Adds an intro services template
> 
> @scottsun94 , after this merges could you also add it to the templates file with a similar config as the Intro to Workspaces template?
> 
> @edoakes , let me know your feedback. I'm not a services expert though so ideally someone on the team could take a pass afterwards to improve the notebook.

Will do a quick review now and happy to do another pass through it after it's merged.

templates/intro-services/README.ipynb
```diff
@@ -0,0 +1,287 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Develop a Serve app locally\n",
+    "\n",
+    " The fastest way to develop a Ray Serve app is locally within the Workspace. A Serve app running within a workspace behaves identically to a Serve app running as a production service, only it does not have a stable DNS name or fault tolerance.\n",
+    "\n",
+    " To get started, create a file called `my_app.py` and fill it with the following skeleton code:\n",
```
edoakes, line 25: nit: to match the convention on other examples we have elsewhere
```suggestion
    " To get started, create a file called `main.py` and fill it with the following skeleton code:\n",
```

templates/intro-services/README.ipynb
```diff
@@ -0,0 +1,287 @@
...
+    "\n",
+    "@serve.deployment\n",
+    "@serve.ingress(app)\n",
+    "class FastAPIDeployment:\n",
+    "    # FastAPI will automatically parse the HTTP request for us.\n",
+    "    @app.get(\"/hello\")\n",
+    "    def say_hello(self, name: str) -> str:\n",
+    "        return f\"Hello {name}!\"\n",
+    "\n",
+    "main = FastAPIDeployment.bind()"
```
edoakes, line 42: (same as above)
```suggestion
    "my_app = FastAPIDeployment.bind()"
```

templates/intro-services/README.ipynb
```diff
@@ -0,0 +1,287 @@
...
+    "import requests\n",
+    "from fastapi import FastAPI\n",
+    "from ray import serve\n",
+    "\n",
+    "app = FastAPI()\n",
+    "\n",
+    "@serve.deployment\n",
+    "@serve.ingress(app)\n",
+    "class FastAPIDeployment:\n",
+    "    # FastAPI will automatically parse the HTTP request for us.\n",
```
edoakes, line 37:
```suggestion
    "    # FastAPI will automatically parse the HTTP request for us.\n",
    "    # Because "name" is specified as an argument, it'll be parsed from the query params of the request.\n",
```

edoakes: Maybe we should link out to fastapi docs?

templates/intro-services/README.ipynb
edoakes: Let's have Gene add this once the logging UI is available?

edoakes (approved):
Looks good only nits, I can iterate on it once `serve deploy` is in.

I'd like to have content somewhere about how to pass arguments to an application because it's a common use case, especially to have different dev. vs. prod args, and it's best to suggest an opinionated way to do it. Do you think a section here makes sense? Or should we have a separate "advanced" notebook? Don't want to have notebook explosion...

templates/intro-services/README.ipynb
ericl: I can add an example.

@akshay-anyscale can you explain to me what I should note here? If I use `print()`, will that not show up in the logs at all?

templates/intro-services/README.ipynb
```diff
@@ -0,0 +1,287 @@
...
+    "print(requests.get(\"http://localhost:8000/hello\", params={\"name\": \"Theodore\"}).json())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Deploy to production as a Service\n",
+    "\n",
+    "In order to enable fault tolerance and expose your app to the public internet, you must \"Deploy\" the application, which will create an Anyscale Service backed by a public load balancer. This Service will run in a separate Ray cluster from the workspace, and will be monitored by the Anyscale control plane to recover on node failures. You will also be able to deploy rolling updates to the service without incurring downtime.\n",
```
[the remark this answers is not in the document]
ericl: Changed to lower case (except when referring to it as "Anyscale Service"), since all upper case would be pretty obnoxious.

templates/intro-services/README.ipynb
ericl: Ok maybe I'll just merge this for now to unblock things.

ericl:
> I'd like to have content somewhere about how to pass arguments to an application because it's a common use case, especially to have different dev. vs. prod args, and it's best to suggest an opinionated way to do it.

Sounds good. Can you let me know some snippets (don't actually know what you have in mind to show)?

templates/intro-services/README.ipynb
akshay-anyscale: it will show up but it won't be annotated with metadata like request id, etc. or follow any formatting (if we want to use json formatting). I agree we can merge and ask Gene to update this

edoakes:
> > I'd like to have content somewhere about how to pass arguments to an application because it's a common use case, especially to have different dev. vs. prod args, and it's best to suggest an opinionated way to do it.
> 
> Sounds good. Can you let me know some snippets (don't actually know what you have in mind to show)?

https://docs.ray.io/en/latest/serve/advanced-guides/app-builder-guide.html#pass-arguments-to-applications

PR #62, opened 2024-02-01
Update stable diffusion template in preparation for serverless

shomilj:
* Change head node type to m5.xlarge so it starts up immediately (we can buffer it)
* Fix a few nits from bug bash
* Removed the example showing how to extract host/token from the stdout because it was a lot of unnecessary text IMO and cluttered the screen (I was looking at three tokens on the screen during the bug bash -- two of which were the example & one was my real token from the stdout).

### DIFF
--- a/configs/serve-stable-diffusion/aws.yaml
+++ b/configs/serve-stable-diffusion/aws.yaml
@@ -1,6 +1,6 @@
 head_node_type:
   name: head_node_type
-  instance_type: g5.xlarge
+  instance_type: m5.xlarge
 worker_node_types:
 - name: gpu_worker
   instance_type: g5.xlarge
--- a/configs/serve-stable-diffusion/gce.yaml
+++ b/configs/serve-stable-diffusion/gce.yaml
@@ -1,7 +1,7 @@
 # n1-standard-8-nvidia-tesla-t4-1  --> 8 CPUs, 1 GPU
 head_node_type:
   name: head_node_type
-  instance_type: n1-standard-8-nvidia-tesla-t4-1
+  instance_type: n1-standard-8
 worker_node_types:
 - name: gpu_worker
   instance_type: n1-standard-8-nvidia-tesla-t4-1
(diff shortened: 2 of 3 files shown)


configs/serve-stable-diffusion/aws.yaml
```diff
@@ -1,6 +1,6 @@
 head_node_type:
   name: head_node_type
-  instance_type: g5.xlarge
+  instance_type: m5.xlarge
```
akshay-anyscale, line 3: should we just make this a mega config? That's what I did for aica - https://github.com/anyscale/templates/tree/main/configs/serve-stable-diffusion-aica

akshay-anyscale: Also we should update the main.py app to use accelerator type resource.

shomilj: Can this workspace template run on any type of GPU, or does it require A10G?

Why would we make it a mega config? I don't see an advantage.

ericl: @akshay-anyscale we shouldn't have any mega configs once the "auto select worker" feature lands

shomilj:
> Also a bit concerned that without g5 pooling in general this isn't the right tradeoff, since workspace start is less painful than a big pause in the middle of serve run waiting for the worker nodes to launch.

We discussed this before, but I think it's just a cost-convenience tradeoff. If we're targeting AIOA customers with this template, they're going to be burning through credits just by reading the workspace template code and understanding how it works (they don't need the GPU up until the point where they actually run their code). I think AICA users probably don't care as much because large companies have $$$.

Let's discuss in the office.

configs/serve-stable-diffusion/aws.yaml
shomilj: +1

akshay-anyscale: we should update the main.py to atleast use accelerator type since that's the pattern we want people to follow

shomilj:
Will merge the change to head node type _after_ buffers are ready

PR #67, opened 2024-02-05
Text Embeddings Template

scottjlee:
Adds README and compute config for a new workspace template for computing text embeddings.

With https://github.com/anyscale/product/pull/26521


templates/text-embeddings/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+![Embeddings Computation Overview](diagram.jpg "Embeddings Computation Overview")
+
+## Running the code
+The provided script will work as-is; no need to modify anything the underlying script, unless you want to customize the logic to meet your specific requirements.
+
+Sample usage:
+```
+# Generate emeddings for a sample single-file parquet file containing 10 rows from the `falcon-refinedweb` HuggingFace dataset, using 1 GPU.
+python ray_embeddings.py \
```
gvspraveen, line 18: where is the python file?

scottjlee: It's in the docker image used for the cluster environment. We want to keep this code private, so it is not included in this public repo.

gvspraveen: can you create a wrapper python file here which takes params and invokes proprietary script in docker image. Seems odd to ask users to directly invoke our python file in docker image

scottjlee: yes, the relevant files (python file, ipynb, sample files, etc) are included in the `default` directory.

gvspraveen (approved):

templates/text-embeddings/README.md
gvspraveen: Awesome thanks for confirming

PR #77, opened 2024-02-15
[FT] Updated the template

shawnanyscale:
- Updated the template with more info on lora finetuning and serving
- Currently users have no easy way of getting fientuend lora model ID.
- TODO: the model ID should be outputted directly in the logs. I will get to this after I come back from vacation

### DIFF
--- a/templates/fine-tune-llm/job_compute_configs/gcp.yaml
+++ b/templates/fine-tune-llm/job_compute_configs/gcp.yaml
@@ -1,8 +1,8 @@
 compute_config:
   allowed_azs:
     - any
-  cloud: my-cloud # You may specify `cloud_id` instead
-  region: any
+  cloud_id: cld_id # You may specify `cloud_id` instead
+  region: us-west1
   head_node_type:
     instance_type: n2-standard-4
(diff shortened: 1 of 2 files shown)


templates/fine-tune-llm/job_compute_configs/gcp.yaml
```diff
@@ -1,8 +1,8 @@
 compute_config:
   allowed_azs:
     - any
-  cloud: my-cloud # You may specify `cloud_id` instead
-  region: any
+  cloud_id: cld_id # You may specify `cloud_id` instead
```
gvspraveen, line 4: hmm this is easy to miss. wonder if we should call out in readme.md that you need to set few required fields. Otherwise if people direct run with this file, they get a error with invalid cloud

gvspraveen (approved):

templates/fine-tune-llm/job_compute_configs/gcp.yaml
gvspraveen: actual i see this mentioned in readme

PR #81, opened 2024-02-16
add automatic generation of readme and github actions

sofianhnaide:
1. This setup automatic generation of README.md for templates using notebooks,
2. Automatically change relative image links with absolute image links (for generated readme's) so it can be viewed from Anyscale console.
3. Add as git pre-commit.
4. Add a github action to block PR's without generated content.


ericl (commented):
nit: should we call the generated readme README.generated.md or generate a big comment on top that warnings the file is auto-generated?

Otherwise it might be confusing if you edit the README and the changes get auto overwritten.

ericl (approved):

ci/auto-generate-readme.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Loop through each readme files
+for readme_file in $readme_files; do
+    # Extract the path of the directory containing the README file, relative to the repository root
+    readme_dir=$(dirname "$readme_file" | sed "s|\.\./templates/||")
+
+    # Check the operating system
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        # macOS system
+        sed -i '' "s|<img src=\"\([^\"http://][^\":/][^\"].*\)\"|<img src=\"${REPO_PREFIX}/${readme_dir}/\1\"|g" "$readme_file"
```
ericl, line 26: lol

sofianhnaide: gpt4 ftw

sofianhnaide:
> nit: should we call the generated readme README.generated.md or generate a big comment on top that warns the file is auto-generated?
> 
> Otherwise it might be confusing if you edit the README and the changes get auto overwritten.

Yes, once all templates are using readme but for now some are still not so that would break them (since the UI is looking for generated and that won't be present).

This will only override templates using notebooks, so if you want to manually edit the readme you should not be using the readme notebook so I feel like this is a consistent state.

ci/auto-generate-readme.sh
ericl: oh I'll have gemini double check it

PR #84, opened 2024-02-21
Make llm serving template serverless

ericl:


templates/intro-services/main.py
```diff
@@ -0,0 +1,16 @@
+import requests
```
ericl, line 1: Add this to make it a bit more friendly out of the box

configs/endpoints_v2/gcp.yaml
```diff
@@ -3,180 +3,5 @@ head_node_type:
...
-  resources:
-    cpu:
-    gpu:
-    memory:
-    object_store_memory:
-    custom_resources:
-      "accelerator_type:A100-80G": 1
-  min_workers: 0
-  max_workers: 100
-gcp_advanced_configurations_json:
```
ericl, line 178: @akshay-anyscale is it ok to remove this?

shomilj (approved):
💯

configs/endpoints_v2/gcp.yaml
akshay-anyscale: I'm assuming these are defaulted in @shomilj 's serverless config? If not, we should keep these

ericl: Shomil can you confirm?

shomilj: I took one more look at the code and realized the A100-80G / A100-40G resources are known to the Autoscaler but we are not making them known to Ray (so my simple tests of asking @ray.remote(accelerator_type="A100-80G") tried to launch them, as expected, but if they actually had launched successfully, then the tasks wouldn't have been scheduled there, since we didn't set RAY_OVERRIDE_RESOURCES).

So I have one more fix to make before we can merge this PR. Doing it now.

shomilj (commented):
https://github.com/anyscale/product/pull/26531 needs to land before this can merge

templates/intro-services/main.py
ericl: It was already there, I just created the file so the user doesn't need to copy paste code from the markdown cell.

akshay-anyscale:
@shomilj is multi-az stuff in serverless defaults

shomilj:
> @shomilj is multi-az stuff in serverless defaults
No, this is something that should be set in the base advanced config if desired. So we should keep these around in the compute configs @ericl:
```
aws:
  TagSpecifications:
    - ResourceType: instance
      Tags:
        - Key: as-feature-enable-multi-az-serve
          Value: "true"
        - Key: as-feature-multi-zone
          Value: "true"
```

gvspraveen:
> No, this is unrelated to serverless - this is something that should be set in the base advanced config if desired.

@shomilj So in this will serverless config only add worker_node_types?

shomilj:
@gvspraveen yes, "serverless" is just "auto_select_worker_config" set to true

ericl:
> No, this is unrelated to serverless - this is something that should be set in the base advanced config if desired. So we should keep these around in the compute configs @ericl:

@shomilj any reason not to enable this by default for all OA configs? It seems strange we have these configs that are required for the top CUJ that are off by default.

shomilj:
https://github.com/anyscale/product/pull/26531 has been merged.

We will likely enable multi-zone by default for OA workloads -- will tackle that as a separate work item this upcoming sprint. So I would add back the aws / gcp advanced config blocks for now, but other than that this is ready to go :)

PR #85, opened 2024-02-23
[finetune-llm] Training configs for 70b full parameter finetuning

amogkam:
Add YAMLs for more Llama 70b finetuning configurations


templates/fine-tune-llm/training_configs/full_param/llama-2-13b-4k-4xg5_12xlarge.yaml
```diff
@@ -8,6 +8,7 @@ train_batch_size_per_device: 8
 eval_batch_size_per_device: 8
 learning_rate: 5e-6
 num_checkpoints_to_keep: 1
+dataset_size_scaling_factor: 10000
```
gvspraveen, line 11: what does this do?

gvspraveen (approved):

templates/fine-tune-llm/training_configs/full_param/llama-2-13b-4k-4xg5_12xlarge.yaml
amogkam: This removes llmforge's restriction on dataset size. This is currently a workaround, but in the future we should remove the default dataset size restriction in llmforge and only enable it for public endpoints. This is being tracked and I will make an issue for this.

PR #89, opened 2024-02-24
Remove custom `custom_llm_gpu` resource specification for `text-embeddings` template

scottjlee:
https://github.com/anyscale/templates/pull/67 created the `text-embeddings` workspace template with a bad config involving `custom_llm_gpu` which prevents satisfying the resource requirement. This PR fixes the issue by removing the unnecessary configuration, since autoscaling works across any of the GPU worker types by default.

### DIFF
--- a/configs/text-embeddings/aws.yaml
+++ b/configs/text-embeddings/aws.yaml
@@ -18,7 +18,7 @@ worker_node_types:
   instance_type: g5.x2large
   resources:
     cpu:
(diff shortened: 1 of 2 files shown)


akshay-anyscale:
can we remove this resource from the application? Not sure what the purpose of this is here if it's assigned to every node type

akshay-anyscale (approved):

PR #93, opened 2024-02-27
LLM Batch Workspace Template

brent-anyscale:
Workspace Template for Batch LLM Workloads.

Includes both a Jupyter Notebook walking through using Ray Data with vLLM as well as a Python and job yaml example.

### DIFF
--- a/templates/batch-llm/README.md
+++ b/templates/batch-llm/README.md
@@ -1,0 +1,42 @@
+# Batch LLM Inference using Anyscale Jobs and vLLM
+
+This template shows you how to:
+* Run a batch LLM inference workload using vLLM and Ray Data
+* Deploy the batch LLM workload as an Anyscale Service
+
+**Note**: This template can be run from this Workspace.
+
+## Review the Jupyter Notebook
+
+## Run as an Anyscale Job
+This Workspace Template includes an example `batch_llm_job.yaml` which can be used to run this batch inference as an Anyscale Job.
+
+[Anyscale Jobs](https://docs.anyscale.com/jobs/get-started) provide key features such as:
+
+* Automated failure handling
+* Automated email alerting
+* Record and persist outputs such as logs
+*
+### Step 1: Review the python code
+
+* Open the `batch_llm.py` file
+* This template is using the prompts in an S3 bucket in AWS. The prompts are in the format:
+```
+I always wanted to be a...
+The best way to learn a new language is...
+```
+* This can be updated to any similarly formatted dataset.
+
+
+## Step 2: Run as an Anyscale Job
+
+To run the as an Ansycale Job from this Workspace, run the following command:
+```bash
+anyscale job submit batch_llm_job.yaml --follow
+```
+
+This will copy the existing python and re-use the [Cluster Environment](https://docs.anyscale.com/configure/dependency-management/cluster-environments) and [Compute Config](https://docs.anyscale.com/configure/compute-configs/overview) to run the job.
+
+## Notes
+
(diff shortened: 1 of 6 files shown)


templates/batch-llm/batch_llm.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+# Initialize Ray with a Runtime Environment - used to set the Environment Variable `HF_TOKEN`
+# This line can be removed if setting the Environment Variable using a Cluster Environment or in the job yaml file.
+ray.init(runtime_env={"env_vars": {"HF_TOKEN": hf_token}})
+
+input_path = "s3://anonymous@air-example-data/prompts.txt"
+ds = ray.data.read_text(
+    input_path,
+    parallelism=10,
```
c21, line 22: Let's remove this line. `parallelism` would be deprecated from Ray Data.

templates/batch-llm/batch_llm.py
```diff
@@ -0,0 +1,54 @@
...
+            "prompt": prompt,
+            "generated_text": generated_text,
+        }
+
+
+ds = ds.map_batches(
+    LLMPredictor,
+    concurrency=10,
+    num_gpus=1,
+    batch_size=32,
```
c21, line 50: Can we keep the same comments as https://github.com/vllm-project/vllm/pull/2696/files ? The comments would be important for users to understand how to set these parameters.

templates/batch-llm/batch_llm.py
```diff
@@ -0,0 +1,54 @@
...
+)
+
+output_path = os.environ.get("ANYSCALE_ARTIFACT_STORAGE") + "/result"
+ds.write_parquet(output_path)
```
c21, line 54: Let's also add the followed same as https://github.com/vllm-project/vllm/pull/2696/files#, and we can comment them out in code. So users can uncomment to try out for a few prompts during testing:

```py
# Peek first 10 results.
# NOTE: This is for local testing and debugging. For production use case,
# one should write full result out as shown below.
# outputs = ds.take(limit=10)
# for output in outputs:
#     prompt = output["prompt"]
#     generated_text = output["generated_text"]
#     print(f"Prompt: {prompt!r}, Generated text: {generated_text!r}")
```

templates/batch-llm/README.md
```diff
@@ -0,0 +1,42 @@
+# Batch LLM Inference using Anyscale Jobs and vLLM
+
+This template shows you how to:
+* Run a batch LLM inference workload using vLLM and Ray Data
```
c21, line 4: Let's add a link for vLLM - https://docs.vllm.ai/en/latest/, and Ray Data - https://docs.ray.io/en/latest/data/data.html?

c21 (commented):
Thank you @brent-anyscale!

templates/batch-llm/README.md
```diff
@@ -0,0 +1,42 @@
+# Batch LLM Inference using Anyscale Jobs and vLLM
+
+This template shows you how to:
+* Run a batch LLM inference workload using [vLLM](https://docs.vllm.ai/en/latest) and [Ray Data](https://docs.ray.io/en/latest/data/data.html)
+* Deploy the batch LLM workload as an Anyscale Service
+
+**Note**: This template can be run from this Workspace.
+
+## Review the Jupyter Notebook
+
```
gvspraveen, line 10: can add a section to run locally in workspace too? 

Basically instruction to run `python batch_llm.py`

gvspraveen: oh wait. is that jupyter note book code?

c21: Make sense, addressed in https://github.com/anyscale/templates/pull/103 .

PR #97, opened 2024-02-28
Upgrade stable diffusion OA template to SDXL

ericl:
Tested manually in notebook. Overall, the quality seems much higher and the speed is pretty similar.

```
image = generate_image("Anyscale logo, as a valentines day card, professional quality art, surrounded by flowers with white envelope background", 640)

filename = "image.png"
with open(filename, "wb") as f:
    f.write(image)

from IPython.display import Image
Image(filename=filename)
```

output

### DIFF
--- a/templates/serve-stable-diffusion/main.py
+++ b/templates/serve-stable-diffusion/main.py
@@ -2,7 +2,7 @@ import logging
 from io import BytesIO
 
 import torch
-from diffusers import EulerDiscreteScheduler, StableDiffusionPipeline
+from diffusers import DiffusionPipeline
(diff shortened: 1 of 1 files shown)


edoakes (approved):
Any Sale 🚀

scottsun94:
When I tested other products, their examples also use refiners to further improve the image quality (like lepton's example: https://github.com/leptonai/examples/blob/main/advanced/sdxl/sdxl.py)

We can probably add those later

ericl:
Yea it's a good point. We should revisit for SD3 (I assume it will have a different set of best practices since it's also a new architecture).

PR #98, opened 2024-02-28
Update template to run serve run inside a terminal and surface GCP in…

akshay-anyscale:
…structions better


ericl (approved):

edoakes (approved):

ericl:
It's interesting this doesn't work well btw, for the intro services template --non-blocking seems to work fine and print "Deployed successfully" prior to existing.

akshay-anyscale:
> It's interesting this doesn't work well btw, for the intro services template --non-blocking seems to work fine and print "Deployed successfully" prior to exiting.

Primarily because the models take much longer to deploy. I think the intro service one just runs on the head node?

ericl:
Hm I see, so it might be some race condition or timeout where we don't print the message in time. Seems possible to fix in the future to improve --no-blocking mode.

PR #99, opened 2024-02-28
Remove spot instances from compute config

akshay-anyscale:


configs/endpoints/aws.yaml
```diff
@@ -20,8 +20,6 @@ worker_node_types:
       "accelerator_type:T4": 1
   min_workers: 0
   max_workers: 100
-  use_spot: true
```
gvspraveen, line 23: hmm. how will customers who want spot instances configure this in template?

akshay-anyscale: they can edit the compute config

gvspraveen (approved):

PR #103, opened 2024-02-28
Add LLM batch inference workspace template

c21:
This is a REDO of https://github.com/anyscale/templates/pull/93 , the reason is to avoid `/` in branch name, pointed in https://github.com/anyscale/product/pull/26646#discussion_r1505379908 .

Tested on workspace https://ci-b1dpzk.anyscale-dev.dev/o/anyscale-internal/workspaces/expwrk_c7vakt4iyfttvmzpjvqnneq5kw/ses_s3llg35wjde9x6j3hezbsgbbrg?workspace-tab=terminal
Full output: https://gist.github.com/c21/5b28f803e0572b73b54dea89a2c54f01


templates/batch-llm/batch_llm.py
```diff
@@ -0,0 +1,78 @@
...
+    num_gpus=1,
+    # Specify the batch size for inference.
+    batch_size=32,
+)
+
+# Write inference output data out as Parquet files to S3.
+# Multiple files would be written to the output destination,
+# and each task would write one or more files separately.
+output_path = os.environ.get("ANYSCALE_ARTIFACT_STORAGE") + "/result"
+ds.write_parquet(output_path)
```
gvspraveen, line 68: can you print log to indicate where result is stored for the user? 

Also will this override past result all the time? should you create new checkpoint for each run (random string or timestamp based)?

templates/batch-llm/batch_llm.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Create a sampling params object.
+sampling_params = SamplingParams(temperature=0)
+
+
+# Create a class to do batch inference.
+class LLMPredictor:
+    def __init__(self):
+        # Create an LLM.
+        self.llm = LLM(model=HF_MODEL)
```
gvspraveen, line 35: how does vllm know where to get model from?

Non blocking for V0: How about custom models. In platform serving lot of times users have finetuned or custom models in s3 they want to run inference on . How do we support that?

templates/batch-llm/batch_llm.py
c21: Good callout. Will do same thing as https://github.com/anyscale/templates/blob/main/templates/fine-tune-llm/train.py#L26 .

templates/batch-llm/batch_llm.py
c21: > how does vllm know where to get model from?

vLLM always loads the model from Hugging Face - https://github.com/vllm-project/vllm/blob/main/docs/source/models/engine_args.rst . The model name has to be Hugging Face model name.

> Non blocking for V0: How about custom models. In platform serving lot of times users have finetuned or custom models in s3 they want to run inference on . How do we support that?

Yes good catch. Discussed offline, for V0, we can just support Hugging Face models. In V1/followup PR, let's tackle the custom/fine-tuned model, and change to use Anyscale vLLM fork.

templates/batch-llm/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+[Anyscale Jobs](https://docs.anyscale.com/jobs/get-started) provide key features such as:
+
+* Automated failure handling
+* Automated email alerting
+* Record and persist outputs such as logs
+
+To run the as an Ansycale Job from this Workspace, run the following command:
+```bash
+anyscale job submit batch_llm_job.yaml --follow
```
akshay-anyscale, line 41: how will it get the vllm dependency?

templates/batch-llm/batch_llm.py
c21: Updated.

templates/batch-llm/batch_llm.py
c21: Discussed with @gvspraveen offline, using Endpoint docker image, which contains Anyscale LLM.

templates/batch-llm/README.md
c21: For V0, decide to not include Anyscale job. Will investigate for V1.

gvspraveen (approved):

PR #106, opened 2024-02-29
Standardize head node type on `m5.2xlarge` for now

shomilj:
* Picking this as a head node that can pull images at a reasonable rate while still being cost-effective enough to buffer (8 CPU's, 32 GiB memory)
* If we do some optimization on image pulling, we can probably drive this down to 4 CPU's, 16 GiB memory nodes (m5.xlarge) assuming Ray/GCS are fine with it for small to medium workloads
* Adding codeowners so that we have some enforcement of this standard (also updated the README)


README.md
```diff
@@ -32,3 +32,14 @@ To add a template:
...
 3. Update the product repo `backend/workspace-templates.yaml` to point to the new template added here after being merged.
+
+## Guidelines for Compute Configurations
+
+All head node configs should be standardized to:
+```yaml
+head_node_type:
+  name: head
+  instance_type: m5.2xlarge
+  resources:
```
ericl, line 43: Can you remove this for now?

shomilj: done

yifeif (approved):

PR #108, opened 2024-02-29
Switch head node of stable diffusion template to CPU only

ericl:
I tested this out manually (including service deploy) and it seems to work well. I specified the worker as min=max=1 so that startup is still fast.

### DIFF
--- a/configs/serve-stable-diffusion/gce.yaml
+++ b/configs/serve-stable-diffusion/gce.yaml
@@ -1,11 +1,10 @@
-# n1-standard-8-nvidia-t4-16gb-1  --> 8 CPUs, 1 GPU
 head_node_type:
   name: head_node_type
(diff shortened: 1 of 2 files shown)


configs/serve-stable-diffusion/gce.yaml
```diff
@@ -1,11 +1,10 @@
...
   name: head_node_type
-  instance_type: n1-standard-8-nvidia-t4-16gb-1
+  instance_type: n1-standard-8
 worker_node_types:
 - name: gpu_worker
   instance_type: n1-standard-8-nvidia-t4-16gb-1
-  min_workers: 0
-  max_workers: 100
+  min_workers: 1
+  max_workers: 1
```
shomilj, line 8: min=max=1 doesn't do anything if auto select is on, btw

shomilj (approved):

PR #111, opened 2024-03-01
Fix some crashes in OA batch inference template due to transformers upgrade

ericl:
- Device seems to be an int now
- Code sample for asserting batch wasn't properly run before

### DIFF
--- a/templates/intro-workspaces/README.ipynb
+++ b/templates/intro-workspaces/README.ipynb
@@ -140,7 +140,7 @@
     "\n",
     "### Using \"Auto-select workers\" mode\n",
     "\n",
-    "To let Ray automatically select what kind of worker nodes to add to the cluster, check the \"Auto-select workers\" box. Ray will add worker nodes as needed to run submitted tasks and actors. In auto mode, you cannot configure worker node types, but the resources panel will show which node types have been launched.\n",
+    "To let Ray automatically select what kind of worker nodes to add to the cluster, check the \"Auto-select workers\" box. Ray will add worker nodes as needed to run submitted tasks and actors. In auto mode, you cannot configure workers, but the resources panel will show which nodes have been launched.\n",
     "\n",
     "We recommend using auto mode if you do not have specific cluster requirements, and are ok with waiting for the autoscaler to add nodes on-demand to the cluster."
    ]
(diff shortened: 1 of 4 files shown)


templates/intro-workspaces/README.ipynb
```diff
@@ -140,7 +140,7 @@
     "\n",
     "### Using \"Auto-select workers\" mode\n",
     "\n",
-    "To let Ray automatically select what kind of worker nodes to add to the cluster, check the \"Auto-select workers\" box. Ray will try to autoscale cluster worker nodes to balance cost and performance. In auto mode, you cannot configure worker node types, but the resources panel will show which node types have been launched.\n",
+    "To let Ray automatically select what kind of worker nodes to add to the cluster, check the \"Auto-select workers\" box. Ray will add worker nodes as needed to run submitted tasks and actors. In auto mode, you cannot configure worker node types, but the resources panel will show which node types have been launched.\n",
```
scottsun94, line 143:
```suggestion
    "To let Ray automatically select what kind of worker nodes to add to the cluster, check the \"Auto-select workers\" box. Ray will add worker nodes as needed to run submitted tasks and actors. In auto mode, you cannot configure workers, but the resources panel will show which nodes have been launched.\n",
```

We don't use "node types" in UI any more

templates/intro-workspaces/README.md
```diff
@@ -89,7 +89,7 @@ To parallelize beyond the resources available to the workspace node, add additio
 
 ### Using "Auto-select workers" mode
 
-To let Ray automatically select what kind of worker nodes to add to the cluster, check the "Auto-select workers" box. Ray will try to autoscale cluster worker nodes to balance cost and performance. In auto mode, you cannot configure worker node types, but the resources panel will show which node types have been launched.
+To let Ray automatically select what kind of worker nodes to add to the cluster, check the "Auto-select workers" box. Ray will add worker nodes as needed to run submitted tasks and actors. In auto mode, you cannot configure worker node types, but the resources panel will show which node types have been launched.
```
scottsun94, line 92:
```suggestion
To let Ray automatically select what kind of worker nodes to add to the cluster, check the "Auto-select workers" box. Ray will add worker nodes as needed to run submitted tasks and actors. In auto mode, you cannot configure worker nodes, but the resources panel will show which nodes have been launched.
```

scottsun94 (approved):

PR #114, opened 2024-03-02
[OA] Improve batch inference example by adding a scaling example

ericl:
- Add a example using 100k rows
- Note I had to manually repartition the dataset in this case (@raulchen will this be fixed for GA?)


scottsun94 (approved):
LGTM.

OOC: do we need to expose the "actor" concept? Or where and how should Ray data users learn the "actor" concept if they've never used ray core before?

ericl:
Yes it's a fair point, so I edited the copy to avoid mentioning actors.

PR #118, opened 2024-03-04
[OA] Add timing info for templates

ericl:
- 10 min for intro workspaces, tune, batch inference
- 5 min for other ones (though haven't timed scientifically)


ericl:
@edoakes should I add "(15 min for GCP)" for all the serving ones? Not sure what else we can do...

edoakes:
> @edoakes should I add "(15 min for GCP)" for all the serving ones? Not sure what else we can do...

Best we can do for now I suppose. We could also tune the rollout strategy to be more aggressive by default to reduce the number of LB updates, but I don't really want to mess around with that right now. Can optimize this based on feedback.

edoakes (approved):

PR #124, opened 2024-03-08
[Endpoints template] Update template to Mistral, provide default adapter

ArturNiederfahrenhorst:
This PR makes it so that the template receives the following main updates:

- Make mistral the default. We are sunsetting support for 7B fine-tuning on product.
- Provide a default adapter from a public mirror
- Adapt the readme to give a quick start with the default adapter


ArturNiederfahrenhorst:
I pushed my HF token to this PR but have since renewed it so no action must be taken here.

templates/endpoints/lora-serve.yaml
```diff
@@ -5,8 +5,10 @@ applications:
     args:
       models: []
       multiplex_models:
-        - "./models/llama/meta-llama--Llama-2-7b-chat-hf_l4-lora.yaml"
-      dynamic_lora_loading_path: s3://my-bucket/my-lora-checkouts/  # <-- change this to a bucket you have access to that contains the loRA checkpoint to load dynamically
+        - "./models/mistral/mistralai--Mistral-7B-Instruct-v0.1_a10g_tp1-lora.yaml"
+      multiplex_lora_adapters: []
+      function_calling_models: []
+      dynamic_lora_loading_path: s3://large-dl-models-mirror/endpoints_template/ # <-- change this to a bucket you have access to that contains the loRA checkpoint to load dynamically
```
akshay-anyscale, line 11: I recommend using the same path as the one above `s3://my-bucket/my-lora-checkouts`

ArturNiederfahrenhorst: @akshay-anyscale That will make it so that there is no default though. With this PR, we want to provide a default. So that users can use this template "as is" without creating a valid LoRA adapter first. Also, if they fail, they should have some way of verifying wether the template works without their modifications.
That's why I've uploaded a sample adapter folder to that s3 mirror

akshay-anyscale: Will customers have access to this bucket though?

ArturNiederfahrenhorst: Galen said it's public!

akshay-anyscale (approved):

PR #126, opened 2024-03-10
update username to strip spaces and make sure it's 5 characters long

kylehh:
1. strip trailing and leading spaces
2. replace intermmediate spaces with _
3. append to make sure len is 5


Aydin-ab:

PR #129, opened 2024-03-11
[LLM] Fixed fine-tuning template username setup

shawnanyscale:


templates/fine-tune-llm/train.py
```diff
@@ -33,7 +33,7 @@ def generate_model_tag(model_id: str) -> str:
     """
     username = os.environ.get("ANYSCALE_USERNAME")
     if username:
-        username = username[:5]
+        username = username.strip().replace(" ", "_")[:5]
```
gvspraveen, line 36: nit: this may add `_` to the end sometimes

```
>>> username = "Kyle abcde"
>>> username = username.strip().replace(" ", "_")[:5]
>>> username
'Kyle_'
```

You can avoid that by following

`username[:5].strip().replace(" ", "_")`

But with my suggestion you could end up with username less than 5 characters sometimes

gvspraveen (approved):

templates/fine-tune-llm/train.py
kylehh: Is it OK to have username less than 5?
I actually submitted a PR #126 to append username to be 5

shawnanyscale: Updated logic so we never have `_` in case there are issues with it and always padded it to be length 5.

I think it's okay to have username less than 5 but just in case it's length 0 for some reason, I always added padding

PR #139, opened 2024-03-14
Add more detailed prompt format documentation

akshay-anyscale:


templates/endpoints/CustomModels.md
```diff
@@ -98,9 +98,58 @@ engine_config:
     bucket_uri: gs://YOUR_BUCKET_NAME/YOUR_MODEL_FOLDER
     extra_files: []
 ```
+### Prompt Format
+A prompt format is used to convert a chat completions API input into a prompt to feed into the LLM engine. The format is a dictionary where the key refers to one of the chat actors and the value is a string template for which to convert the content of the message into a string. Each message in the API input is formated into a string and these strings are assembled together to form the final prompt.
+
+The string template should include the `{instruction}` keyword, which will be replaced with message content from the ChatCompletions API.
+
+For example, if a user sends the following message for llama2-7b-chat-hf ([prompt format](continuous_batching/meta-llama--Llama-2-7b-chat-hf.yaml#L27-L33)):
```
gvspraveen, line 106: `continuous_batching/meta-llama--Llama-2-7b-chat-hf.yaml#L27-L33`

dead link?

akshay-anyscale: fixed

templates/endpoints/CustomModels.md
```diff
@@ -98,9 +98,58 @@ engine_config:
...
+
+What is the capital of France? [/INST] The capital of France is Paris. </s><s>[INST] What about Germany? [/INST]
+```
+
+#### Schema
+
+The following keys are supported:
+* `system` - The system message. This is a message inserted at the beginning of the prompt to provide instructions for the LLM.
+* `assistant` - The assistant message. These messages are from the past turns of the assistant as defined in the list of messages provided in the ChatCompletions API.
+* `trailing_assistant` - The new assistant message. This is the message that the assistant will send in the current turn as generated by the LLM.
```
kouroshHakha, line 143:
```suggestion
* `trailing_assistant` - The special characters that will be added to the end of the prompt before sending it to the LLM for generation. This often includes special characters that put the LLM into assitant mode (granted that model has been trained to support such special keywords). For example [vicuna](https://huggingface.co/TheBloke/vicuna-13B-v1.5-GGML) has `Assistant: ` as a special tag that can go here. 
```

kouroshHakha (commented):
Just one small modification about trailing_assistant.

Yard1:
@kouroshHakha can you also open a PR to modify the models/README.md in rayllm so they are consistent?

akshay-anyscale:
> Just one small modification about trailing_assistant.

updated

kouroshHakha:
> @kouroshHakha can you also open a PR to modify the models/README.md in rayllm so they are consistent?
@akshay-anyscale can you open up a PR against rayllm as well?

akshay-anyscale:
@kouroshHakha can you approve?

kouroshHakha (approved):

PR #142, opened 2024-03-15
[BYOD] Automation to build and snapshot template docker image

ftimyo:
https://docs.google.com/document/d/1DZ40-bFhQLR4b1pmkgyLtiA1brH46RsOJZhlrBD3WMc/edit

This PR create a script to automate the process of the SOP in the doc To run this script, one should have following tools installed 1) bazelisk # plan to remove this dependency soon
2) git # plan to remove this dependency soon
3) docker
4) aws

Before running the script, one should setup the credential from anyscale-dev-product account on AWS.

This script should run on both devbox and laptop, and devbox is recommended. follow this task to track progress of this project https://anyscale1.atlassian.net/browse/AE-529


build_byod_snapshot.sh
hckuo, line 1: 's/build_byod_snapshot.sh/build_image_snapshot.sh`

build_byod_snapshot.sh
```diff
@@ -0,0 +1,319 @@
+#!/bin/bash
+
+set -uo pipefail
+
+# planned improvement will be commented in the following task
+# https://anyscale1.atlassian.net/browse/AE-529
+
```
hckuo, line 7: can you also put dependencies in of this shell script in the comment?
e.g., docker, 100G of disk space etc...

build_byod_snapshot.sh
```diff
@@ -0,0 +1,319 @@
...
+    print_info "cleanup local registry"
+    docker stop registry
+    docker rm registry
+}
+
+init_setup() {
+    parse_args "$@"
+    export SNAPSHOTS_DIR="$(mktemp -d -p /tmp)"
+    export TMP_REPO_DIR="$(mktemp -d -p /tmp)"
+    export CRANE_URL="https://github.com/google/go-containerregistry/releases/download/v0.15.1/go-containerregistry_Linux_x86_64.tar.gz"
```
hckuo, line 130: parametrize this version.

hckuo, line 130: make these global and declared at the top with the default values so that advanced users can override them.

build_byod_snapshot.sh
```diff
@@ -0,0 +1,319 @@
...
+            aws s3 cp $SNAPSHOTS_DIR/$s s3://$bucket/$path/$s;
+        fi
+    done
+}
+
+push_snapshots() {
+    print_info "Attempt pushing snapshots to premerge"
+    (
+        # assume role in subshell
+        assume_role "arn:aws:iam::830883877497:role/buildkite-deploy-to-premerge"
```
hckuo, line 277: Can you also parameterize these ARN? e.g., PROD_ARN, STAGING_ARN at the top?

build_byod_snapshot.sh
```diff
@@ -0,0 +1,319 @@
...
+        upload_snapshots_to anyscale-dev-organization-data-us-west-2
```
hckuo, line 278: bucket name too

build_byod_snapshot.sh
```diff
@@ -0,0 +1,319 @@
...
+    print_info "Starting local Docker registry"
+    docker run -d -p $DOCKER_REGISTRY_PORT:5000 --restart always --name registry registry:2
+    print_info "Local Docker registry started, listening port $DOCKER_REGISTRY_PORT"
+}
+
+setup_tool_repo() {
+    pushd $TMP_REPO_DIR
+    print_info "Cloning ray-project/ray.git"
+    # TODO @timy remove this once dockerfile is placed in anyscale/templates repo
+    git clone git@github.com:ray-project/ray.git
```
hckuo, line 64: why do we need ray project here?

build_byod_snapshot.sh
ftimyo: will parametize bucket names as global variables.

build_byod_snapshot.sh
ftimyo: This is temporary, so I can test the script. The docker image file is not avaialbe in template yet.

build_byod_snapshot.sh
```diff
@@ -0,0 +1,319 @@
...
+    echo "Install crane"
+    pause
+    if [ $? -eq 1 ]; then
+        setup_crane
+    else
+        echo "Skip install crane"
+    fi
+}
+
+build_docker_image() {
```
hckuo, line 167: maybe we should just change the API to take a dockerfile and a context dir

build_byod_snapshot.sh
```diff
@@ -0,0 +1,319 @@
...
+    # TODO @timy remove this once dockerfile is placed in anyscale/templates repo
+    git clone git@github.com:ray-project/ray.git
+    print_info "Cloning anyscale/product.git"
+    git clone git@github.com:anyscale/product.git
+    popd # TMP_REPO_DIR
+}
+
+setup_crane() {
+    print_info "Downloading crane from ${CRANE_URL}"
+    curl -sL "${CRANE_URL}" | tar xzf - -C /tmp/
```
csivanich, line 72:
```suggestion
    curl -sSfL "${CRANE_URL}" | tar xzf - -C /tmp/
```
Nit: to report error status codes to pipefail via rc=1

csivanich:
All looks good to me - however I have some questions about whether this should be in this public repo or not. If so, we should at least decouple the magic strings/ids from the source.

ftimyo:
After collecting and applying all comments from this PR, I'll create a new one in private repo and close this one.

PR #144, opened 2024-03-18
[Template] Removed multi-zone config for training jobs

shawnanyscale:
- Removed multi-zone config for training jobs
- thread: https://anyscaleteam.slack.com/archives/CKRJJNN4E/p1710635724823809

### DIFF
--- a/templates/fine-tune-llm/job_compute_configs/gcp.yaml
+++ b/templates/fine-tune-llm/job_compute_configs/gcp.yaml
@@ -22,15 +22,9 @@ compute_config:
           a2-highgpu-8g-nvidia-a100-40gb-8: 8
   gcp_advanced_configurations_json:
     instance_properties:
-      labels:
-        as-feature-multi-zone: "true"
       disks:
         - boot: true
           auto_delete: true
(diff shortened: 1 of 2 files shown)


templates/fine-tune-llm/job_compute_configs/gcp.yaml
```diff
@@ -22,15 +22,9 @@ compute_config:
...
   gcp_advanced_configurations_json:
     instance_properties:
-      labels:
-        as-feature-multi-zone: "true"
       disks:
         - boot: true
           auto_delete: true
           initialize_params:
             disk_size_gb: 250
-runtime_env:
```
galenhwang, line 32: reason for removal?

galenhwang (approved):

templates/fine-tune-llm/job_compute_configs/gcp.yaml
shawnanyscale: Those are auto-populated by llmforge and no longer needed. For instance, they are not needed on the aws file

PR #148, opened 2024-03-20
[OA][template] Finetune stable-diffusion (dreambooth)

justinvyu:
Adds a dreambooth stable diffusion finetuning template for OA.

[The original dreambooth template](https://docs.ray.io/en/latest/train/examples/pytorch/dreambooth_finetuning.html) re-implemented a [huggingface accelerate example](https://huggingface.co/docs/diffusers/en/training/dreambooth).

Additions of the new template:
* Uses the huggingface example rather than re-implementing it -- this reduces our maintenance burden in the future to upgrade to SD3. Let's just take the latest version of the diffusers example rather than build a separate version in parallel.
    * This also makes it easier to add other methods that customers are considering like textual inversion, etc. No need to rebuild all those examples.
* Fine-tunes SDXL instead of SD1.5. ("Using the latest and greatest models.")
* Shifts the focus of the template away from understanding the training code + learning about Ray Data where it's not really needed, to showing that it is easy to launch a distributed training job from a user's existing script. Fewer concepts to learn when getting started.
* The data size is very small for stable diffusion fine-tuning, which makes Ray Data not as important to showcase.

Removals:
* No longer uses ray data for "batch inference" to generate images in parallel to construct the regularization dataset when using a prior preservation loss. It is now done serially (the original script logic).

This takes inspiration from the [Modal template](https://modal.com/docs/examples/dreambooth_app), which writes very minimal code and leverages the existing huggingface example. Modal just focuses on showing how easy it is to take a huggingface accelerate script and run it on a cluster. We can show the same thing with Ray Train + Anyscale.


templates/fine-tune-stable-diffusion/train.py
```diff
@@ -0,0 +1,53 @@
...
+
+def launch_finetuning(input_args=None) -> ray.train.Result:
+    args = parse_args(input_args=input_args)
+
+    trainer = TorchTrainer(
+        train_fn_per_worker,
+        # Pass arguments from the driver to the `config` dict of the `train_fn_per_worker`
+        train_loop_config={"args": args},
+        scaling_config=ray.train.ScalingConfig(
+            num_workers=1,
```
matthewdeng, line 38: Does it work if we increase `num_workers`?

justinvyu: upped to 4

ericl (commented):
Tried this out, pretty cool. A couple comments:
1. Since we're trying to de-emphasize NFS, can you move these storage paths to use $ANYSCALE_ARTIFACT_STORAGE instead of NFS?
2. The notebook has a lot of spam in the cell outputs, can you clear the outputs before saving them? If there are particular outputs to highlight you can include those as screenshots explicitly in the markdown of the cell.

ericl:
Also, I think the last cell "print("\n".join(finetuned_images))
display(*[Image(filename=image_path, width=250) for image_path in finetuned_images])" is repeated twice.

ericl:
```
huggingface_hub.utils._validators.HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': '/mnt/cluster_storage/checkpoint-final'. Use `repo_type` argument if needed.
```

When I ran it, it didn't seem to succeed saving the checkpoint also.

ericl:
Try it here: https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_px58enxhtvjh5vwysyvh4p8x2p/ses_wgm6f872qhegth6wiahvz99nkv?workspace-tab=vscode&command-history-section=application_logs&search=checkpoint

justinvyu:
> Since we're trying to de-emphasize NFS, can you move these storage paths to use $ANYSCALE_ARTIFACT_STORAGE instead of NFS?

A lot of the usage is pretty awkward to replace with S3: the HF script just reads from `SUBJECT_IMAGES_DIR` as a directory, and `HF_HOME` is very useful to set so that worker startup is quicker since it can skip the model download time. Is it ok to trade-off showing NFS usage for a better user experience?

Is the recommendation to never use NFS for performance reasons? Is it still okay as a place to share some lightweight files between nodes?

>  Also, I think the last cell "print("\n".join(finetuned_images))
display(*[Image(filename=image_path, width=250) for image_path in finetuned_images])" is repeated twice.

The last 2 cells should be showing before vs. after fine-tuning images.

justinvyu:
Also, another bug at the moment is that autoscaling w/ `accelerator_type` is broken and is waiting on https://github.com/ray-project/ray/pull/44225 to be included in nightly.

ericl:
Unfortunately, we have a plan to remove NFS in the next coming sprints for
AIOA clouds, due to high cost to maintain.

justinvyu:
@ericl Got it, is it ok to assume S3 for now? Will we be adding a GCP hosted cloud any time soon?

ericl:
Maybe you can use pyarrow to open the filesystem whether it's GCP or AWS? We are currently onboarding to both platforms.

```
import pyarrow.fs
import os

fs, path = pyarrow.fs.FileSystem.from_uri(os.environ["ANYSCALE_ARTIFACT_STORAGE"])
print(fs, path)

fs.get_file_info(path + "/my-final-checkpoint")
```

templates/fine-tune-stable-diffusion/README.ipynb
```diff
@@ -0,0 +1,390 @@
...
+    "from ray.train.torch import TorchTrainer\n",
+    "\n",
+    "from train_dreambooth_lora_sdxl import main\n",
+    "from utils import download_from_cloud, upload_to_cloud\n",
+    "\n",
+    "\n",
+    "# Set environment variables across the entire cluster.\n",
+    "ray.init(\n",
+    "    runtime_env={\n",
+    "        \"env_vars\": {\"WANDB_API_KEY\": os.environ.get(\"WANDB_API_KEY\")}\n",
```
ericl, line 261: This actually crashes if you don't have WANDB_API_KEY set. Make it optional / comment out this block by default?

templates/fine-tune-stable-diffusion/README.ipynb
```diff
@@ -0,0 +1,390 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Launch the training.\n",
+    "trainer.fit()"
```
ericl, line 296: Combine this with cell above, it's easy to miss.

templates/fine-tune-stable-diffusion/README.ipynb
```diff
@@ -0,0 +1,390 @@
...
+    "        upload_to_cloud(local_path=final_checkpoint_path, cloud_uri=MODEL_CHECKPOINT_PATH)\n",
+    "\n",
+    "\n",
+    "trainer = TorchTrainer(\n",
+    "    train_fn_per_worker,\n",
+    "    # Pass command line arguments from the driver to the `config` dict of the `train_fn_per_worker`\n",
+    "    train_loop_config={\"args\": TRAINING_ARGS},\n",
+    "    scaling_config=ray.train.ScalingConfig(\n",
+    "        # Do data parallel training with A10G GPU workers\n",
+    "        num_workers=4, use_gpu=True, accelerator_type=\"A10G\"\n",
```
ericl, line 284: This is a lot of workers for a tiny job, why not 1 or 2 at most?

ericl:
@justinvyu I'm still having issues after commenting out the broken wandb line, any idea?

```
RayTaskError(TypeError): ray::_Inner.train() (pid=2620, ip=10.0.40.86, actor_id=1d725aa6167d331c9a4e7e3204000000, repr=TorchTrainer)
  File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/tune/trainable/trainable.py", line 334, in train
    raise skipped from exception_cause(skipped)
  File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/train/_internal/utils.py", line 53, in check_for_failure
    ray.get(object_ref)
ray.exceptions.RayTaskError(TypeError): ray::_RayTrainWorker__execute.get_next() (pid=2656, ip=10.0.13.128, actor_id=d62f6cf6e383bebea1201b0a04000000, repr=<ray.train._internal.worker_group.RayTrainWorker object at 0x7f964c517c10>)
  File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/train/_internal/worker_group.py", line 33, in __execute
    raise skipped from exception_cause(skipped)
  File "/home/ray/anaconda3/lib/python3.9/site-packages/ray/train/_internal/utils.py", line 169, in discard_return_wrapper
    train_func(*args, **kwargs)
  File "/tmp/ipykernel_1246/2651988845.py", line 15, in train_fn_per_worker
TypeError: cannot unpack non-iterable NoneType object

The above exception was the direct cause of the following exception:

TrainingFailedError                       Traceback (most recent call last)
Cell In[11], line 2
      1 # Launch the training.
----> 2 trainer.fit()

File ~/anaconda3/lib/python3.9/site-packages/ray/train/base_trainer.py:638, in BaseTrainer.fit(self)
    634 result = result_grid[0]
    635 if result.error:
...
    641 return result
```

ericl:
This workspace btw https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_lwx3xcrpwc1m4kvmc8icdyp5wn/ses_89hgq68tkwrige3qawxvpabb65?workspace-tab=vscode

templates/fine-tune-stable-diffusion/README.ipynb
justinvyu: The original example uses an effective batch size of 4 by accumulating a batch size of 1 on a single device. Now, I just scale out the cluster to 4 gpus and don't do gradient accumulation to make the training run faster.

Finetuning requires around 100 steps to get decent results, which actually takes quite a bit of time with fewer workers. The current setup takes around 8 minutes.

justinvyu:
@ericl Oh oops, I made some changes without doing another round of testing -- there was a bug in the train function. I'll make the fixes and post a link to a workspace with a working version.

justinvyu:
Here is a workspace with the fixed template: https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_s5yeb3qu6wheis18ksujmx2ttb/ses_nja945defkghjxpz2yi39956sc?workspace-tab=vscode

templates/fine-tune-stable-diffusion/README.ipynb
```diff
@@ -0,0 +1,392 @@
...
+    "        print(\"Final checkpoint has been uploaded to: \", MODEL_CHECKPOINT_PATH)\n",
+    "\n",
+    "\n",
+    "trainer = TorchTrainer(\n",
+    "    train_fn_per_worker,\n",
+    "    # Pass command line arguments to the `config` dict of the `train_fn_per_worker`\n",
+    "    train_loop_config={\"args\": TRAINING_ARGS},\n",
+    "    scaling_config=ray.train.ScalingConfig(\n",
+    "        # Do data parallel training with A10G GPU workers\n",
+    "        num_workers=4, use_gpu=True, accelerator_type=\"A10G\"\n",
```
justinvyu, line 292: Problem with this:
* A10G is not available on GCP clouds, only L4
* No easy way to detect which cloud is being used.
* Not possible to specify minimum GPU memory requirement.

Possible solutions:
* Janky way of detecting cloud by checking the `ANYSCALE_ARTIFACT_STORAGE`
* Add a helper that does [requests instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html) and determines the cloud that way.
* Ask the user to uncomment code that switches the accelerator type to L4 if they're running on GCP.

ericl: Yeah just add a helper in utils to `get_A10G_class_gpu()` or similar.

templates/fine-tune-stable-diffusion/README.ipynb
```diff
@@ -0,0 +1,392 @@
...
+    "        prompts=PROMPTS, args=TRAINING_ARGS, model_checkpoint_path=MODEL_CHECKPOINT_PATH\n",
+    "    )\n",
+    "])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Images generated with the base model"
```
ericl, line 335: This is a little confusing, how about just having a cell to display the finedtuned images, and a comment you can uncomment on how to display images for the base model?

Users will not read carefully and might think these are the fine-tuned images.

ericl (approved):
Latest changes look good, just a few minor comments + try it out on GCP.

templates/fine-tune-stable-diffusion/README.ipynb
justinvyu: Good point, I'll comment out the base model generation

justinvyu:
[Template also works on GCP.](https://console.anyscale-staging.com/v2/cld_6zjdsbn3kbphvk3luhyrd3eg6e/workspaces/expwrk_sqrxcvdduijht4cdc6u7n11j2f/ses_8keka1jlmx9h353szrtsudhkyl?workspace-tab=vscode&command-history-section=application_logs)

@ericl Could you merge?

ericl:
Merged. Similar to the other PR, @justinvyu can you follow up to make PRs to add it to the product repo for OA home and also endpoints-doc repo?

justinvyu:
Ok done, ptal!

PR #156, opened 2024-03-27
[OA] Fixes for Batch Inference Basics template

scottjlee:
Address feedback / fixes from dogfooding batch LLM template:
- Fix bug with `text` vs `item` column from `from_items()` call
- Use Mistral model by default, so users are not required to supply HF token
- Use single quotes instead of triple quotes for prompt data
- ~Move scaling sections to after step 4~ vLLM requires GPUs, so need to talk about GPUs in the toy setup as well.
- Clean up title + headers
- Add accelerator type (A10G or L4)

Alongside https://github.com/anyscale/product/pull/27262


templates/batch-llm/README.ipynb
```diff
@@ -262,7 +212,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Apply batch inference for all input data with the Ray Data [`map_batches`](https://docs.ray.io/en/latest/data/api/doc/ray.data.Dataset.map_batches.html) method. Here, you can easily configure Ray Data to scale the number of LLM instances and compute (number of GPUs to use)."
+    "## Scaling with GPUs\n",
+    "\n",
+    "Apply batch inference for all input data with the Ray Data [`map_batches`](https://docs.ray.io/en/latest/data/api/doc/ray.data.Dataset.map_batches.html) method. When using vLLM, LLM instances require GPUs; here, we will demonstrate how to configure Ray Data to scale the number of LLM instances and GPUs needed.\n",
+    "\n",
+    "To use GPUs for inference in the Workspace, we can specify `num_gpus` and `concurrency` in the `ds.map_batches()` call below to indicate the number of LLM instances and the number of GPUs per LLM instance, respectively. For example, with `concurrency=4` and `num_gpus=1`, we have 4 LLM instances, each using 1 GPU, so we need 4 GPUs total."
```
scottjlee, line 219: note, since vLLM requires GPUs, i had to put this section before the "scaling to larger dataset," since we will need GPUs for even the toy setup.

templates/batch-llm/README.ipynb
```diff
@@ -262,7 +212,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Apply batch inference for all input data with the Ray Data [`map_batches`](https://docs.ray.io/en/latest/data/api/doc/ray.data.Dataset.map_batches.html) method. Here, you can easily configure Ray Data to scale the number of LLM instances and compute (number of GPUs to use)."
+    "## Scaling with GPUs\n",
```
ericl, line 215: Since this is under step 4, use "###".

templates/batch-llm/README.ipynb
```diff
@@ -371,7 +387,25 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Summary\n",
+    "## Submitting an Anyscale Job\n",
```
ericl, line 390: Remove this for now, the jobs tutorial isn't ready yet. When it is, we can link to that instead of repeating the same content in each template.

templates/batch-llm/README.ipynb
```diff
@@ -274,28 +228,89 @@
...
+   "outputs": [],
+   "source": [
+    "ds.take_all()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Scaling to a larger dataset\n",
```
ericl, line 271: Similar comment below, if these are all under section 4 they need to be one level deeper as headings.

templates/batch-llm/README.ipynb
```diff
@@ -274,28 +228,89 @@
     "ds = ds.map_batches(\n",
     "    LLMPredictor,\n",
     "    # Set the concurrency to the number of LLM instances.\n",
-    "    concurrency=num_llm_instances,\n",
+    "    concurrency=4,\n",
     "    # Specify the number of GPUs required per LLM instance.\n",
-    "    num_gpus=num_gpus_per_instance,\n",
+    "    num_gpus=1,\n",
```
ericl, line 233: When I ran this, I got

"""
    raise ValueError(
ValueError: Bfloat16 is only supported on GPUs with compute capability of at least 8.0. Your Tesla T4 GPU has compute capability 7.5. You can use float16 instead by explicitly setting the`dtype` flag in CLI, for example: --dtype=half.
"""

Similar to in https://github.com/anyscale/templates/pull/148 I think you need to set `accelerator_type: A10G` and/or make a function that returns A10G or L4 depending on AWS or GCP.

templates/batch-llm/README.ipynb
ericl: Also, this will be "ray job submit" within workspaces.

ericl:
Please ping when it runs correctly in OA, still doesn't work for me

templates/batch-llm/README.ipynb
scottjlee: ah my bad, i was testing this on a custom workspace with A10s already configured, that makes sense. Adding a similar function as [#148](https://github.com/anyscale/templates/pull/148) which gets A10G/L4 depending on the cloud platform.

scottjlee:
@ericl tested on OA workspace ([link](https://console.anyscale-staging.com/v2/cld_kqz1dtput5gh1t8rrh2apdy466/workspaces/expwrk_4tq8ctq6nn4acbqbr11msjt1vw/ses_16sxywau2m15t7x96uze4kf2hw?workspace-tab=vscode)) with serverless, ready for another look. Thanks!

ericl (approved):
Nice, works e2e now, thanks!

Btw, I couldn't access the workspace you linked probably b/c it wasn't in the staging dogfood org, for sharing workspaces you probably want to use the "Try new UI" function in staging.

PR #159, opened 2024-03-29
Updates for `batch-llm` template

scottjlee:
- Once we use the latest endpoints AICA image for the `batch-llm` template, we should not install `vllm`/`transformer`, as we do not want to overwrite vllm already installed in the image. We can therefore remove the package installation step from the template.
- Also add section on Jobs page.

Tested on [OA workspace](https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_qx7l6vn83391kp5gcmcab71phb/ses_kdb3s7ixbmsjdw6w6qecx77bma?workspace-tab=vscode)
With https://github.com/anyscale/product/pull/27375


ericl (approved):

Wendi-anyscale (approved):
LGTM!

PR #160, opened 2024-03-30
Update `text-embeddings` template

scottjlee:
- Instead of including the embeddings computation logic in the docker image, we instead copy the logic from `runtime` (https://github.com/anyscale/runtime/pull/507) into a local file used in the workspace. This way, the user can easily customize the logic for their own use case.
- Add section for Jobs, linking to the Intro to Jobs page.

Tested on [OA workspace](https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_uw29mztdz8z5az79vmqitcraft/ses_9v1fkddrtlauygzndkreuues31?workspace-tab=vscode)

### DIFF
--- a/templates/text-embeddings/README.md
+++ b/templates/text-embeddings/README.md
@@ -14,7 +14,29 @@ For a Python script version of the `.ipynb` notebook used for the workspace temp
 **Note:** This tutorial is run within a workspace. Please overview the `Introduction to Workspaces` template first before this tutorial.
 
 ## Step 1: Setup model defaults
-First, let's import the dependencies we will use in this template.
+
+First, install additional required dependencies using `pip`.
+
+
+```python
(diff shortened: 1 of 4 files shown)


templates/text-embeddings/README.md
```diff
@@ -14,7 +14,29 @@ For a Python script version of the `.ipynb` notebook used for the workspace temp
 **Note:** This tutorial is run within a workspace. Please overview the `Introduction to Workspaces` template first before this tutorial.
 
 ## Step 1: Setup model defaults
-First, let's import the dependencies we will use in this template.
+
+First, install additional required dependencies using `pip`.
+
+
+```python
+!pip install -q optimum[onnxruntime-gpu] langchain && echo 'Install complete!
```
scottjlee, line 22: these are not part of the AICA image, so manually need to install.

ericl (approved):

PR #163, opened 2024-04-03
[OA-676] Add back second part of job tutorial

ericl:
- Add back programmatic submission now that CLI is updated


templates/intro-jobs/README.ipynb
```diff
@@ -97,6 +97,24 @@
...
+    "You can also submit jobs from other machines, using the Anyscale CLI. \n",
+    "\n",
+    "Copy ``main.py`` to an empty folder on your **laptop**, and then run the following on your **laptop** to try this out:\n",
+    "\n",
+    "```bash\n",
+    "# Make sure we have anyscale CLI installed.\n",
+    "$ pip install -U anyscale\n",
+    "\n",
+    "# Note: will package all contents of your current working dir into the job deps.\n",
+    "$ anyscale job submit --wait -- python main.py\n",
```
edoakes, line 113: this has to pass `--working-dir=.`, it's currently only defaulted when in workspaces. i don't remember why we decided on that

ericl: Oh, shall we change that? Seems like a good default to have.

ericl: I filed https://anyscale1.atlassian.net/browse/OA-725 to unblock this PR.

edoakes: The problem with setting it by default is we'd also need to have a `--no-working-dir` option, else there's no way to disable uploading. I'd be OK with it if we add that option as well.

ericl:
OK to merge, but offline flow won't work until https://github.com/anyscale/product/pull/27459/files is live.

ericl:
This is ready for review again.

templates/intro-jobs/README.md
```diff
@@ -58,10 +61,29 @@ You should see the job state and its output on the overview page.
...
+
+```bash
+# Make sure we have anyscale CLI installed.
+$ pip install -U anyscale
+
+# Note: outside of workspaces, you must specify required files via --working-dir.
+$ anyscale job submit --working-dir=. --wait -- python main.py
+```
+
+Jobs submitted externally will run with the Anyscale default compute config and dependencies. To override these settings, use the ``--config-file``, ``--image-uri``, or ``--containerfile`` flags.
```
ericl, line 80: Btw, @edoakes is the default the serverless config?

edoakes: yes it is

edoakes (approved):

PR #172, opened 2024-04-10
Add DBRX endpoints template

Yard1:
Adds a DBRX endpoints template using OSS vLLM.


Yard1:
Marking as draft as I need to test it first

akshay-anyscale (approved):

PR #173, opened 2024-04-15
Adding in name to prevent Job job-job-... name

omatthew98:
When we submit jobs without an explicit name the prefix `job-` is prepended to the name of the workspace. In this case, the name of the Intro to Jobs workspace already starts with the `job-` prefix, so we submit a job with the name `job-job-...`. This PR changes the template to submit a job with the name `my-job` preventing the email from saying "Anyscale Job job-job-...".

Before:
image

After:
image


scottjlee (approved):

tson1111 (approved):
LGTM, thanks for the fix!

edoakes (approved):
thx -- good attention to detail

PR #176, opened 2024-04-18
Update `batch-llm` template

scottjlee:
Addresses comments/fixes from dogfooding the template, including:
- New method for prompting user to enter Hugging Face token via VSCode overlay. This is only triggered for LLaMA models, and user no longer need to set any variable manually in the notebook.
Screenshot at Apr 17 21-30-44

- Fix forever looping output text with model-specific prompt instructions.
- Update `main.py` script with these changes
- Various copy and formatting fixes

Tested on [workspace](https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_v2qjb7mz8q1dvm6eft2gtwvbqh/ses_rxfvq5yc8f6amh6ezcg6i6lea6?workspace-tab=vscode)


templates/intro-jobs/README.md
```diff
@@ -48,7 +48,7 @@ The following cell should also run to completion within a few minutes and print
 ```python
 # Second, submit the Ray app for execution on a new Ray cluster.
 # The execution will be managed by Anyscale Jobs.
-!ray job submit --wait -- python main.py
+!ray job submit --name my-job --wait -- python main.py
```
scottjlee, line 51: unrelated to the rest of the PR, but fixes lint for https://github.com/anyscale/templates/pull/173

templates/batch-llm/README.ipynb
```diff
@@ -80,16 +83,35 @@
...
    "outputs": [],
    "source": [
-    "# Set the Hugging Face token. Replace the following with your token.\n",
-    "HF_TOKEN = \"<REPLACE_WITH_YOUR_HUGGING_FACE_USER_TOKEN>\"\n",
-    "# Set to the model that you wish to use from the preceding list.\n",
-    "# Note that using the Llama models will require a Hugging Face token to be set.\n",
-    "HF_MODEL = \"mistralai/Mistral-7B-Instruct-v0.1\"\n",
+    "# Set to the name of the Hugging Face model that you wish to use from the preceding list.\n",
+    "# Note that using the Llama models will prompt you to set your Hugging Face user token.\n",
+    "HF_MODEL = \"meta-llama/Llama-2-7b-chat-hf\"\n",
```
c21, line 88: why we switch the model from Mistral to llama? This would require users to provide tokens, and harder for them to start.

templates/batch-llm/main.py
```diff
@@ -75,15 +97,25 @@ def __call__(self, batch: Dict[str, np.ndarray]) -> Dict[str, list]:
...
     # Specify the number of GPUs required per LLM instance.
     num_gpus=num_gpus_per_instance,
-    # Specify the batch size for inference. Set the batch size to as large possible without running out of memory.
-    # If you encounter CUDA out-of-memory errors, decreasing batch_size may help.
-    batch_size=10,
+    # Specify the batch size for inference. Set the batch size to as large
+    # as possible without running out of memory.
+    # If you encounter CUDA out-of-memory errors, decreasing
+    # batch_size may help.
+    batch_size=5,
```
c21, line 114: Any reason to reduce batch size?

c21 (approved):
Looks good with two Qs.

templates/batch-llm/main.py
scottjlee: this was the switch to prevent stragglers where we would have some actors end up with more batches than others, setting the batch size to number of batches evenly divisible by number of actors (100 rows, 5 rows per batch, 4 actors -> each actor gets 5 batches of 5 rows)

templates/batch-llm/README.ipynb
scottjlee: ah good call, let me change this back to mistral, thanks

templates/batch-llm/main.py
c21: Got it. After fixing the `would have some actors end up with more batches than others` as we planned in https://docs.google.com/document/d/1vw5XIPqHK90O-7098kY4F3aDnuFxKfrpTPC8w2WG9mg/edit?usp=drive_open&ouid=104454598218723057493 , then we don't have this limitation, right?

scottjlee: i think this is a slightly different issue than needing to call `repartition()` before `map_batches()`. In that case, the issue was that there as only 1 input block produced from reading 1 file, so we need to call `repartition(4)` to create enough blocks. This batch size reduction has a different effect, where we want to prevent some actors waiting around for others to finish. 

To illustrate with an example, with 100 input rows and 4 actors, and a batch size of 10:

Actor 1: (0, 10), (40, 50), (80, 90)
Actor 2: (10, 20), (50, 60), (90, 100)
Actor 3: (20, 30), (60, 70)
Actor 4: (30, 40), (70, 80)

In the above example, Actor 3 and 4 will need to wait for actor 1 and 2 to finish their third batches, since there aren't enough batches to go around. If we adjust the batch size to 5, we will have 20 total batches, with each actor getting 5 even batches.

PR #178, opened 2024-04-19
Add LLaMA 3 on `batch-llm` template

scottjlee:
Update `batch-llm` workspace template to support LLaMA 3 8B model.

Also update the login method for getting Hugging Face token, so that we only do it if the model requires it. We first check this by loading the AutoConfig for the model, which will fail if a token is required. In this case, we trigger the authentication via VS Code overlay.

Tested on [workspace](https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_jiit4ufssay8fkw1wfvv5tp744/ses_531agcuc51q5m2l3nvrzqjiwcj?workspace-tab=vscode)

### DIFF
--- a/templates/batch-llm/README.md
+++ b/templates/batch-llm/README.md
@@ -136,7 +136,7 @@ model_name_to_input_prompt_format = {
     "google/gemma-7b-it": "<start_of_turn>model\n{}<end_of_turn>\n",
     "mlabonne/NeuralHermes-2.5-Mistral-7B": "<|im_start|>system\nYou are a helpful assistant that will complete the sentence in the given input prompt.<|im_end|>\n<|im_start|>user{}<|im_end|>\n<|im_start|>assistant",
     "meta-llama/Meta-Llama-3-8B-Instruct": (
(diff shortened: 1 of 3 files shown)


templates/batch-llm/README.md
```diff
@@ -128,18 +135,32 @@ model_name_to_input_prompt_format = {
...
+    construct the input prompt for the model."""
+    prompt_format = model_name_to_input_prompt_format.get(HF_MODEL)
+    if prompt_format:
+        row[text_column] = prompt_format.format(row[text_column])
     return row
 
 ds = ds.map(construct_input_prompt, fn_kwargs={"text_column": "item"})
 ```
 
+So far, we have defined two operations of the Dataset (`from_items()`, `map()`), but have not executed the Dataset yet and don't see any results. Why is that?
```
scottjlee, line 156: moved this up from previous section. since this is the first point where we call `map()` without executing the dataset, user may be confused why there is nothing running.

c21 (approved):
Thanks for adding this!

PR #179, opened 2024-04-19
Added token setup to example commands

shawnanyscale:
- Mistral recently gated their repos such that HF token is needed


shrekris-anyscale (approved):
Do we need to make a config change similar to https://github.com/anyscale/ray-llm/commit/91c199eca72732b466ae3cd132127278759d485b?

akshay-anyscale (approved):

PR #180, opened 2024-04-23
Update Hugging Face token fetching in `batch-llm` template

scottjlee:
Refactor the Hugging Face token fetching mechanism used in `batch-llm` template so that the common utility function can also be used in the Python script used for job submission.

Tested on [workspace](https://console.anyscale-staging.com/v2/cld_kvedZWag2qA8i5BjxUevf5i7/workspaces/expwrk_ad2uga9wmqi393sjkguu11nnh7/ses_ww5pylrj6fqsfimlfn48g4mtwq?workspace-tab=vscode)

### DIFF
--- a/templates/batch-llm/util/utils.py
+++ b/templates/batch-llm/util/utils.py
@@ -3,6 +3,11 @@ import random
 import requests
 import string
 
+import huggingface_hub
+from transformers import AutoConfig
+
+HF_TOKEN_CACHE_PATH = "/home/ray/.cache/huggingface/token"
+
 
 def generate_output_path(output_path_prefix: str, model_id: str) -> str:
     """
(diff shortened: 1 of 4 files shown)


templates/batch-llm/util/utils.py
```diff
@@ -36,3 +41,35 @@ def get_a10g_or_equivalent_accelerator_type() -> str:
...
+    """Prompts the user for Hugging Face token if required by the model.
+    Returns the token as a string. If a token is not required by the model,
+    returns an empty string."""
+
+    try:
+        # Try loading model config to check if Hugging Face token is required.
+        AutoConfig.from_pretrained(hf_model)
+        # If config is loaded successfully, no need for token.
+        return ""
+    except OSError:
```
omatthew98, line 56: Are there other OSErrors that could come from the above? Should we check the message of the exception to make sure it matches?

omatthew98 (approved):
One small q, but lgtm!

templates/batch-llm/util/utils.py
scottjlee: good point, added a check for the specific error type

PR #182, opened 2024-04-24
add non-blocking option and note to accept terms for HF access to models

angelinalg:
Closes JIRA https://anyscale1.atlassian.net/browse/OA-994
cc: @scottsun94 @akshay-anyscale @shawnanyscale

### DIFF
--- a/templates/endpoints_v2/README.ipynb
+++ b/templates/endpoints_v2/README.ipynb
@@ -24,7 +24,7 @@
     "\n",
     "Currently tensor parallelism defaults to 1 if not specified.\n",
     "\n",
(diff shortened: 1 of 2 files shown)


scottsun94 (approved):
LGTM. The hf token part is still easy to miss though.

angelinalg:
What do you suggest, @scottsun94?

angelinalg:
Here's a commit with the Note bolded.

scottsun94:
> Here's a commit with the Note bolded.

I think we can start with this and see how it goes, but currently all the models listed there are either from llama or mistral families. That means that you have to set it up if you want to use them.

PR #185, opened 2024-04-25
Adding in try_create_dir=False to unblock OA release

omatthew98:
Due to how the Anyscale Artifact Storage IAM policy is configured and how the PyArrow s3fs tries to create directories we were getting errors from these templates when we tried to write to the Anyscale Artifact Storage. More context on that issue [in slack here](https://anyscaleteam.slack.com/archives/C0695Q8TZSN/p1713828788038729). 

We plan to fix the behavior on the Ray Data side to by default not try to create directories for s3 as this call is not necessary, but for now this patch in the templates will resolve the issue while we wait for the next Ray release (will happen after the OA release).

### DIFF
--- a/templates/batch-llm/README.ipynb
+++ b/templates/batch-llm/README.ipynb
@@ -383,7 +383,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "ds.write_parquet(output_path)\n",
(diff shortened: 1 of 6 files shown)


c21 (approved):
LGTM

scottjlee (approved):

PR #186, opened 2024-04-26
Enable PLD in Mistral

comaniac:
Enable prompt lookup decoding for Mistral models.
Reference: https://github.com/anyscale/ray-llm/pull/1561, https://github.com/anyscale/ray-llm/pull/1589

@kouroshHakha @Yard1 @csivanich please help review this PR.


Yard1 (approved):
lgtm but we should make sure to document the lack of logprob support somewhere

PR #189, opened 2024-04-30
Use accelerator type and serverless config for sdxl template on OA

scottsun94:
1. Use accelerator type in deployment definition
2. switch to "auto-select worker nodes". Not sure why we use fixed cluster. It doesn't seem necessary.


scottsun94:
Hmm. The pre-commit checks ran and passed. Not sure why it still failed

Screenshot 2024-04-29 at 8 43 23 PM

configs/serve-stable-diffusion/aws.yaml
```diff
@@ -1,10 +1,20 @@
...
   name: head
   instance_type: m5.2xlarge
+  resources:
+    cpu: 0
+
+auto_select_worker_config: true
+
+# This will be ignored since auto_select_worker_config is enabled,
+# but must be present to pass backend validation.
+# TODO (shomilj): Remove once validation is fixed.
```
scottsun94, line 11: @shomilj are they still required?

scottsun94:
Bump. It seems that @akshay-anyscale is the owner of this template, please help review when you get time. Thanks!

templates/serve-stable-diffusion/main.py
```diff
@@ -49,8 +49,13 @@ async def generate(self, prompt: str, img_size: int = 512):
 
 
 @serve.deployment(
-    ray_actor_options={"num_gpus": 1, "num_cpus": 1},  # Set the number of GPUs and CPUs required for each model replica.
-    max_concurrent_queries=2,   # Maximum number of queries that are sent to a replica of this deployment without receiving a response.
+    ray_actor_options={
+        # Set the number of GPUs and CPUs required for each model replica.
+        "num_gpus": 1,
+        "num_cpus": 1,
+        "accelerator_type": "A10G" # Set accelerator type based on GPU type to use (T4, A10G, L4, V100, A100-40G or A100-80G)
```
shomilj, line 56:
```suggestion
        # Set this to your desired GPU type (e.g. T4, A10G, L4, V100, A100-40G, A100-80G).
        "accelerator_type": "A10G"
```

configs/serve-stable-diffusion/aws.yaml
shomilj: you can set:
worker_node_types: []

shomilj (approved):
LGTM, please test this exact configuration (pairing compute configs + code changes) before merging.

scottsun94:
> flags:
  allow-cross-zone-autoscaling: true

Just to confirm, setting this flag to be true is the same as unchecking this.
- I ran into an issue during test when not checking this. The g5 instance type is not supported in the zone of head node.
- It launches node normally after unchecking it
Screenshot 2024-04-30 at 10 26 53 PM

https://github.com/anyscale/templates/assets/9677264/0084eb34-da7b-4605-b1c0-9e6fc0c8193e

scottsun94:
The template works fine, but I have several questions about how the serve behavior:
- why are the APIIngress replica and the SDXL model replica not packed on the same node? 
- controller and proxy are still scheduled onto head node when cpu=0 is set on head node. Not sure if this is desired

https://github.com/anyscale/templates/assets/9677264/fc61090d-4b64-4c16-91df-1cec9dc9d4f6

edoakes:
> The template works fine, but I have several questions about how the serve behavior:
> 
> * why are the APIIngress replica and the SDXL model replica not packed on the same node? (it seems not necessary to spin up a separate cpu worker node)
> * controller and proxy are still scheduled onto head node when cpu=0 is set on head node. Not sure if this is desired
> 
>  Screen.Recording.2024-04-30.at.10.12.14.PM.copy.mov

Hmm I'm not sure about the scheduling issue. This could be a timing issue where the APIIngress gets created first and we don't batch them together so we first schedule the CPU actor, CPU node is requested, then we schedule the GPU actor, and GPU node is requested. The compaction logic that @zcin added should address it by migrating the CPU replica, but I believe it's still flagged off outside of endpoints. Let me see if we can turn it on.

edoakes:
> > The template works fine, but I have several questions about how the serve behavior:
> > 
> > * why are the APIIngress replica and the SDXL model replica not packed on the same node? (it seems not necessary to spin up a separate cpu worker node)
> > * controller and proxy are still scheduled onto head node when cpu=0 is set on head node. Not sure if this is desired
> > 
> > Screen.Recording.2024-04-30.at.10.12.14.PM.copy.mov
> 
> Hmm I'm not sure about the scheduling issue. This could be a timing issue where the APIIngress gets created first and we don't batch them together so we first schedule the CPU actor, CPU node is requested, then we schedule the GPU actor, and GPU node is requested. The compaction logic that @zcin added should address it by migrating the CPU replica, but I believe it's still flagged off outside of endpoints. Let me see if we can turn it on.

It is expected that the proxy and controller still run on the head node.

scottsun94:
@edoakes Shall I merge first or wait for the compaction flag?

edoakes:
> @edoakes Shall I merge first or wait for the compaction flag?

No need to wait let's merge

PR #190, opened 2024-04-30
Refine Hugging Face token auth in `batch-llm` template

scottjlee:
This PR refines the logic in `batch-llm` workspace template for authenticatin Hugging Face token from user input. 
- Instead of relying on environment variable read from the file cached by Hugging Face, we copy the token to a local file in the working directory of the workspace template, which can be read by the job we launch later.
- Also replace the model config loading with a simpler HTTP request to their model registry, which serves as a check to see if the model requires a token.
- Move the token authentication to its own cell to denote importance

Also remove text related to tensor parallelism, since it is confusing for most non-advanced users without documentation (we currently don't have good documentation around this)

### DIFF
--- a/templates/batch-llm/README.ipynb
+++ b/templates/batch-llm/README.ipynb
@@ -494,7 +494,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.9.19"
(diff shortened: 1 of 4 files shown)


omatthew98 (approved):
Lgtm!

It seems like the python version was changed from 3.11 to 3.9? If that is important maybe we should remove that but seems like it probably doesn't matter for this.

scottjlee:
> It seems like the python version was changed from 3.11 to 3.9? If that is important maybe we should remove that but seems like it probably doesn't matter for this.

thanks, just reverted it

PR #191, opened 2024-05-01
Stable diffusion pretraining template

marwan116:


configs/stable-diffusion-pretraining/gce.yaml
```diff
@@ -0,0 +1,20 @@
+head_node_type:
+  name: head
+  instance_type: n2-standard-8
+  resources:
+    cpu: 0
+auto_select_worker_config: false
```
shomilj, line 6: Why can this not use serverless?

marwan116: My desired goal is a single GPU, with no autoscaling, a very simple set up to ensure the smoothest UX for anyone trying out the template.

shomilj: Can't you just enforce a single GPU via the application?

marwan116: I see - I am primarily optimizing for "time to run notebook"

but I see your point in not consuming GPU units and I am willing to try it your way - perhaps will make the change and then test it out - if it is still a reasonable experience will keep as is, otherwise will escalate

shomilj: The upscaling speed will be much faster in production, where we are buffering warm A10G's. If you are using the slim image as your base, upscaling an A10G should be on the order of < 20 seconds.

marwan116: ok nice - yes I am hoping to make use of the slim image. Sounds promising

marwan116: @shomilj what about `allow-cross-zone-autoscaling` should I be setting this to true or false? what considerations should I keep in mind  here? - I am thinking cross-zone autoscaling entails a delay so perhaps best to keep it false ?

configs/stable-diffusion-pretraining/aws.yaml
```diff
@@ -0,0 +1,20 @@
...
+  instance_type: g5.4xlarge
+  resources:
+    cpu:
+    gpu:
+    memory:
+    object_store_memory:
+    custom_resources:
+      "accelerator_type:A10G": 1
+  min_workers: 0
+  max_workers: 1
```
shomilj, line 18:
```suggestion
worker_node_types: []
```

configs/stable-diffusion-pretraining/aws.yaml
```diff
@@ -0,0 +1,20 @@
...
+flags:
+  allow-cross-zone-autoscaling: false
```
shomilj, line 20:
```suggestion
flags:
  allow-cross-zone-autoscaling: true
```

configs/stable-diffusion-pretraining/gce.yaml
```diff
@@ -0,0 +1,20 @@
...
+    cpu:
+    gpu:
+    memory:
+    object_store_memory:
+    custom_resources:
+      "accelerator_type:T4": 1
+  min_workers: 0
+  max_workers: 1
+flags:
+  allow-cross-zone-autoscaling: false
```
shomilj, line 20:
```suggestion
flags:
  allow-cross-zone-autoscaling: true
```

configs/stable-diffusion-pretraining/gce.yaml
```diff
@@ -0,0 +1,20 @@
...
+  instance_type: n1-standard-8-nvidia-t4-16gb-1
+  resources:
+    cpu:
+    gpu:
+    memory:
+    object_store_memory:
+    custom_resources:
+      "accelerator_type:T4": 1
+  min_workers: 0
+  max_workers: 1
```
shomilj, line 18:
```suggestion
worker_node_types: []
```

shomilj (approved):
approving compute config changes

PR #196, opened 2024-05-03
[parallel-experiments] use s3fs for s3 storage

matthewdeng:
Use `s3fs` instead of the default `pyarrow` `S3FileSystem`.

`ANYSCALE_ARTIFACT_STORAGE` uses a shared bucket.  The `pyarrow` implementation makes (unneeded) requests that require permissions that are (intentionally) not given.

A long term fix will be to remove these calls from the `pyarrow` imeplementation.

### DIFF
--- a/templates/intro-tune/filesystem_utils.py
+++ b/templates/intro-tune/filesystem_utils.py
@@ -1,0 +1,22 @@
+from s3fs import S3FileSystem
+import pyarrow.fs
+from typing import Optional, Tuple
+
+
+def get_path_and_fs(storage_path: str) -> Tuple[str, Optional[pyarrow.fs.FileSystem]]:
+    """Get the final storage path and corresponding filesystem, if applicable.
+
+    For S3, this will use s3fs.
+    For others, this will use the default filesystem(None).
+    """
+    s3_prefix = "s3://"
+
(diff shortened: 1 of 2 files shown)


justinvyu (commented):
TY, changes look good.

Can we also update the `load_data` call to pass in a fixed path so that all nodes will only download once? Like `load_data("/mnt/local_storage/cifar_data")`?

justinvyu (approved):

lanbochen-anyscale (approved):

templates/intro-tune/filesystem_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+def get_path_and_fs(storage_path: str) -> Tuple[str, Optional[pyarrow.fs.FileSystem]]:
+    """Get the final storage path and corresponding filesystem, if applicable.
+
+    For S3, this will use s3fs.
+    For others, this will use the default filesystem(None).
+    """
+    s3_prefix = "s3://"
+
+    if storage_path.startswith(s3_prefix):
```
akshay-anyscale, line 14: does this work for GCP?

matthewdeng: Yep this is customization specific to S3 + Anyscale Hosted.

PR #198, opened 2024-05-06
Update fine-tuning template

ArturNiederfahrenhorst:
This PR makes the following modifications:

- Introduce llama 3 8B/70B fine-tuning configs
- Introduce Mixtral fine-tuning config
- Introduce Mistral fine-tuning config
- Remove llama 2 13b fine-tuning configs. Folks should use llama 3 8B or maybe Mistral, let's not get them confused.
- Move supported models to FAQ so that people who come to this template try the correct models *only*
- Renames deepspeed configs to reflect their settings rather than the model they can/should be used for. This should keep people from having to look and understand what's inside when expanding on this template.
- Updated to the README, most notably adding some more FAQs to better support OA users who want to expand on the template but should be self-sufficient

All existing and new compute configs have been tested by me so that they don't OOM and are optimized.


templates/fine-tune-llm_v2/README.md
```diff
@@ -1,72 +1,65 @@
...
 
-# Launch a full-param fine-tuning job for Llama 7b with 16 A10s
-!python main.py training_configs/full_param/llama-2-7b-512-16xa10.yaml
+# Launch a full-param fine-tuning job for Llama 3 8B with 16 A10s
+!python main.py training_configs/full_param/llama-3-8b.yaml
 
-# Launch a LoRA fine-tuning job for Llama 7b with 16 A10s
-# !python main.py training_configs/lora/llama-2-7b-512-16xa10.yaml
+# Launch a LoRA fine-tuning job for Llama 3 8B with 16 A10s
+# !python main.py training_configs/lora/llama-3-8b.yaml
```
ArturNiederfahrenhorst, line 32: I've tested these lines manaully

templates/fine-tune-llm_v2/README.ipynb
```diff
@@ -4,33 +4,28 @@
    "cell_type": "markdown",
```
ArturNiederfahrenhorst, line 4: This reflects the README.md file and does not have to be reviewed

akshay-anyscale (approved):

templates/fine-tune-llm_v2/README.md
```diff
@@ -85,4 +78,56 @@ You can edit the values, such as `context_length`, `num_epoch`, `train_batch_siz
 In addition, the deepspeed configs are provided in case you would
 like to customize them.
 
+### What's the full list of supported models?
+
+The following models can currently be fine-tuned with this template.
```
kouroshHakha, line 83: Give a general guideline, say in general any model that is compatible with the architecture of the mentioned base models can be finetuned for both LoRA and full parameter using the same configs as the base models. 

This is a growing list but it includes the following models:

templates/fine-tune-llm_v2/README.md
```diff
@@ -85,4 +78,56 @@ You can edit the values, such as `context_length`, `num_epoch`, `train_batch_siz
...
+- meta-llama/Meta-Llama-3-70B
+- meta-llama/Meta-Llama-3-70B-Instruct
+
+### What if I want to use another dataset? 
+
+The training configs provided in this template all train on the GSM8k which requires a context length of 512 tokens.
+How to ensure the correct format is described in https://docs.endpoints.anyscale.com/fine-tuning/dataset-prep.
+You can replace the s3 buckets in the training configs in this template with paths to your own dataset.
+
+### I have the right model, context length and everything. Can I optimize compute cost?
```
kouroshHakha, line 103: love this. Let's add more color to it

templates/fine-tune-llm_v2/README.md
```diff
@@ -85,4 +78,56 @@ You can edit the values, such as `context_length`, `num_epoch`, `train_batch_siz
...
+
+Optimizing your fine-tuning runs for compute cost is a non-trivial problem.
+The default configs in this template require the following compute:
+Llama-3-8B and Mistral require 16 A10Gs. Llama-3-70B and Mixtral require 32 A10Gs.
+If you want different compute, we *suggest* the following workflow to find a suitable configuration:
+
+* Select some model, context length, fine-tuning technique (LoRA or full-paramter), etc. that suit the problem you want to solve
```
kouroshHakha, line 110: emphasize the context length required for the problem setting is important and it impacts how much batch size you can fit. Batchsize is determined for the worst case and you want to make sure at largest batch size you don't oom. Mention that if they have small number of data points that required huge context lenghts, see if you can remove those data points to reduce context length varations.

templates/fine-tune-llm_v2/README.md
```diff
@@ -85,4 +78,56 @@ You can edit the values, such as `context_length`, `num_epoch`, `train_batch_siz
...
+* Start off with compute that you think will give you good flops/$. If you are not sure, here is a rough guideline:
+    * g5 nodes for high availability
+    * p4d/p4de nodes for lower availability but better flops/$
+    * Anything more modern if you have the means of acquiring them
+* Do some iterations of trial and error on batch-size and deepspeed settings to fit the workload while keeping other settings fixed
+    * Start with batch sizes of 1
+    * Use deepspeed stage 3
+    * Try to use deepspeed offloading only if it reduces the minimum number of instances you have to use
```
kouroshHakha, line 118: Mention that deepspeed offloading makes the training slow, but it allows larger batch sizes to be achieved because of more relaxed GRAM foot-print.

kouroshHakha: Mention that it is recommended to do the finetuning on the same machine if possible. The GPU to GPU communication across machines is very expensive compared to the memory savings it could provide. So doing training on A100s may end up cheaper than doing multi-node A10s if availablity is not an issue.

templates/fine-tune-llm_v2/training_configs/full_param/mistral-7b.yaml
```diff
@@ -1,20 +1,18 @@
...
+model_id: mistralai/Mistral-7B-Instruct-v0.1 # <-- change this to the model you want to fine-tune
 train_path: s3://air-example-data/gsm8k/train.jsonl # <-- change this to the path to your training data
 valid_path: s3://air-example-data/gsm8k/test.jsonl # <-- change this to the path to your validation data. This is optional
-context_length: 4096 # <-- change this to the context length you want to use
+context_length: 512 # <-- change this to the context length you want to use
 num_devices: 16 # <-- change this to total number of GPUs that you want to use
 num_epochs: 10 # <-- change this to the number of epochs that you want to train for
 train_batch_size_per_device: 8
-eval_batch_size_per_device: 8
+eval_batch_size_per_device: 16
```
kouroshHakha, line 8: We usually never deviate from train and eval batch size being the same, doing so just guranteed to not oom during evaluation, because you don't have backward pass.

templates/fine-tune-llm_v2/README.md
```diff
@@ -85,4 +78,56 @@ You can edit the values, such as `context_length`, `num_epoch`, `train_batch_siz
...
+We do not guarantee that this will give you optimal settings, but have found this workflow to be helpful ourselves in the past.
+
+### How can I get even more control?
+
+This template fine-tunes with Anyscale's library `llmforge`, which uses [DeepSpeed](https://github.com/microsoft/DeepSpeed) and [Ray Train](https://docs.ray.io/en/latest/train/train.html) for distributed training.
+You can study main.py to find out how we call the `lmforge dev finetune` API with a YAML that specifies the fine-tuning workload.
+You can call `lmforge dev finetune` yourself and gain control by modifying the training config YAMLs in this template.
+For anything that goes beyond using `llmforge`, you can build your own fine-tuning stack on Anyscale.
+
+### What's with the `main` file that is created when I fine-tuning?
```
kouroshHakha, line 131: Add a question about considerations about LoRA. how to modify it? What to expect, etc.

kouroshHakha (commented):
Awesome clean up. A few asks for some improvements:

templates/fine-tune-llm_v2/training_configs/full_param/mistral-7b.yaml
```diff
@@ -11,10 +11,8 @@ num_checkpoints_to_keep: 1
 dataset_size_scaling_factor: 10000
 output_dir: /mnt/local_storage
 deepspeed:
-  config_path: deepspeed_configs/zero_3_llama_2_13b.json
+  config_path: deepspeed_configs/zero_3_offload_optim+param.json
 dataset_size_scaling_factor: 10000 # internal flag. No need to change
 flash_attention_2: true
```
kouroshHakha, line 16: why remove the `trainer_resouces` ?

kouroshHakha: This ensures the trainer worker picks the large g5 nodes so that it doesn't OOM during checkpointing.

kouroshHakha (commented):
One Q:

templates/fine-tune-llm_v2/training_configs/full_param/mistral-7b.yaml
ArturNiederfahrenhorst: I believe this was there because the config was previously for Llama 2 13 with 4k context length. I renamed it and reused it for Mistral now (llama 3 8B works without this line as well)

kouroshHakha: But the one for llama-3-8B still has this memory requirement. It may work because you are giving it the gpu resources you need. If you don’t specify mem requirements it may get smallest g5 node.

ArturNiederfahrenhorst: I don't see the line you are referring to. Looking at the Llama 3 8B configs I don't see this.

But you are still right, I did another round of testing and have updated the compute configs.
I trusted the worker and trainer resources there but started OOMing on super small instances.
I have updated some configs based on using 16 (or 32) small nodes that did not OOM.

kouroshHakha (approved):

kouroshHakha:
Good to merge from my end.

PR #199, opened 2024-05-06
Delete old ft templates

ArturNiederfahrenhorst:
As communicated via Slack, the GPT-J and llama 2 fine-tuning templates are out of date and will not be part of OA.
Poeple should only use the other, updated fine-tuning templates.


akshay-anyscale:
Also need to remove from https://github.com/anyscale/product/blob/master/backend/workspace-templates.yaml

akshay-anyscale (approved):

akshay-anyscale:
can you also remove the compute configs?

PR #203, opened 2024-05-07
LLM router basic template

anmscale:
Implementing:
* basic flow for data labeling with gpt4 as a judge
* basic flow for evaluation

To do:
* add 1-2 simple baslines (frequency, BoW classifier)
* Iterate on the explanation and overall story

In a following commit:
* finetune llm router


templates/llm-router/README.ipynb
```diff
@@ -0,0 +1,1038 @@
...
+    "More concurrently, let us assume that closed models have perfect a quality (5/5 score). our goal is to reduce cost significantly (say by 50%) while maintaining a high overal quality (4.8/5 score).\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Implementation overview\n",
+    "\n",
+    "### 1) Generate labeled data\n",
```
kouroshHakha, line 32: ### 1) Generate labeled data
The label for such classifier does not exist, and we explore one way to generate such data for a specific use-case.

templates/llm-router/README.ipynb
```diff
@@ -0,0 +1,1038 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Background\n",
```
kouroshHakha, line 7: A few open question:

What are some concerns about runtime of injecting a router at the for-front of this decision? How does the latency of this model impact the architecture choice of it?

templates/llm-router/README.ipynb
```diff
@@ -0,0 +1,1038 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Implementation overview\n",
+    "\n",
+    "### 1) Generate labeled data\n",
+    "\n",
+    "### 2) Finetune a router model on the labeled data\n",
```
kouroshHakha, line 34:
```suggestion
    "### 2) Create a router model via fine-tuning small LLMs on labeled data\n",
```

templates/llm-router/README.ipynb
kouroshHakha: add a glimpse of explanation to each section ^

templates/llm-router/README.ipynb
kouroshHakha, line 1: Let's make sure we silent the warnings given by pandas or other libraries in the code. We want to keep the code outputs clean.
image

kouroshHakha: Let's move all the imports to the top (first code block after Approach is finished)

First import all the libraries that are external, then in a separate code block import all the stuff that are helper methods / classes within this repo.

kouroshHakha: Let's add a markdown explanation on what each code block would do, for example here:
image


or here 
image

templates/llm-router/README.ipynb
```diff
@@ -0,0 +1,1038 @@
...
+    "dataset_df[\"label\"] = dataset_df.index.map(labels)\n",
+    "dataset_df.to_json(\"assets/labeled_test_dataset.json\")\n",
+    "dataset_df.head()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Assume we labeled the full dataset (~120K samples)\n",
```
kouroshHakha, line 747: Some where mention that the notebook is running this data collection at small scale. 

For larger scale you need to do x y z, also this is how much it'll cost, etc.

templates/llm-router/README.ipynb
```diff
@@ -0,0 +1,1038 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "\n",
+    "dataset_df = pd.read_json(\n",
+    "    \"/mnt/user_storage/templates_data/labeled_full_dataset.jsonl\",\n",
```
kouroshHakha, line 762: We need to put dataset file in remote storage. (public s3 bucket or github lfs, etc) See the other finetuning end to end example already on the endpoints docs. https://docs.endpoints.anyscale.com/examples/e2e-finetune-and-serve-example

templates/llm-router/README.ipynb
```diff
@@ -0,0 +1,1038 @@
...
+   "metadata": {},
+   "source": [
+    "# Finetune an LLM classifier"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Format data in Anyscale finetuning format"
```
kouroshHakha, line 902: Link to what Anyscale's finetuning format looks like

templates/llm-router/data_utils.py
```diff
@@ -0,0 +1,178 @@
+import re
+import matplotlib.pyplot as plt
+from collections import Counter
+import pandas as pd
+
+
+def preprocess_nectar(df, model, model_name):
+    """
```
kouroshHakha, line 8: Universal comment for all methods:
Always use type hinting (for both inputs and outputs) + docstring format should be:

```
"""<one line explanation>

<more detailed explanation>

Args:
     foo: explain foo
     bar: explain bar

Returns:
     explain returned items.
"""
```

kouroshHakha: For this function in particular let's add more explanation in the docstring on what the preprocessing is doing? Like: Filter to include only the first turn, good-natured responses, and responses that contain the specified model

templates/llm-router/data_utils.py
```diff
@@ -0,0 +1,178 @@
...
+            ret.append({"role": "assistant", "content": turn})
+    return ret
+
+
+def prepare_llm_queries(dataset_df, system_message=None):
+    """Prepare queries for using LLM endpoints"""
+    queries = {}
+    for pidx, row in dataset_df.to_dict(orient="index").items():
+        prompt = row["prompt"]
+        if type(prompt) == str:
```
kouroshHakha, line 66:
```suggestion
        if isinstance(prompt, str):
```

templates/llm-router/data_utils.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    Split the dataset into training and validation sets.
+    """
+    validation_df = dataset_df.sample(n=validation_size, random_state=42)
+    train_df = dataset_df.drop(validation_df.index)
+
+    return train_df, validation_df
+
+
+def visualize_label_distribution(dataset_df, key):
```
kouroshHakha, line 122: Let's move all visualization methods to a different module.

templates/llm-router/evaluation_metrics.py
```diff
@@ -0,0 +1,67 @@
...
+        oss_scores[:, np.newaxis],
+        np.ones_like(model_predictions) * 5,
+    )
+    average_score = chosen_scores.mean(axis=0)
+    score_auc = sklearn.metrics.auc(routing_percentage, average_score)
+
+    return average_score, routing_percentage, score_auc
+
+
+def plot_quality_cost_curve(
```
kouroshHakha, line 25: Move visualization into another module (maybe viz.py)

templates/llm-router/online_inference.py
```diff
@@ -0,0 +1,81 @@
...
+import ray
+from typing import Dict, Any, List
+import copy
+import openai
+import time
+import ray
+
+
+@ray.remote(num_cpus=0)
+def get_llm_response(
```
kouroshHakha, line 12: Any mechanisms to guard against rate limits?

templates/llm-router/online_inference.py
```diff
@@ -0,0 +1,81 @@
...
+            )
+            return (pidx, response.choices[0].message.content)
+        except Exception as e:
+            print(f"Exception: {e}")
+            time.sleep(retry_interval)  # default is per-minute rate limits
+            retry_count += 1
+    return (pidx, "")
+
+
+def generate_batch_responses(
```
kouroshHakha, line 45: definitely make the docstring for this beefy.

kouroshHakha (commented):
Overall I think it's going in the right direction, Let's continue. It's half baked towards the end so I didn't put feedback there.

A few things I like so far that I think we should make standard for all the templates:

- README.ipynb should not have implementation details in code. All the code implementation should be abstracted in a separate module / function that gets imported and just used. 

- We should use diagrams when they cut the need for more words.

templates/llm-router/README.ipynb
GokuMohandas, line 1: This looks good so far but I'm really interested in how you will explain/show the training workload in the notebook. I'm sure you'll explain parts of the config, how the new classifier head works, applying the template (and again during inference). I'm also excited for the serving part. I think you should have a very small cost analysis too!

Small errors:
- "running them on a very number of queries"
- "the closed always" --> "the closed model"
- "use GPT-4 as a representative of closed models" --> "use GPT to represent closed models"
- "More concurrently" --> "More concretely"

configs/llm-router/aws.yaml
```diff
@@ -0,0 +1,19 @@
+head_node_type:
+  name: head
+  instance_type: p4de.24xlarge
```
shomilj, line 3: A100 head nodes are not available in Hosted OA - if this template will be exposed in Hosted OA, can we use the serverless config (just directly request a GPU resource of A100-80G and allow the autoscaler to upscale it?)

If we are not planning on exposing this through OA, then it doesn't matter as much. But it's still better practice to run workloads on workers and use cheap CPU nodes for development.

anmscale: would you consider A10 as cheap gpu? I have enabled training on A10s and launched jobs successfully with it, so I can update this config

configs/llm-router/aws.yaml
```diff
@@ -0,0 +1,19 @@
+head_node_type:
+  name: head
+  instance_type: g5.48xlarge
+  resources:
+    cpu: 8
+
+auto_select_worker_config: true
```
shomilj, line 7: can you delete everything from this line below? none of it should be needed for single node

configs/llm-router/aws.yaml
```diff
@@ -0,0 +1,19 @@
+head_node_type:
+  name: head
+  instance_type: g5.48xlarge
+  resources:
+    cpu: 8
```
shomilj, line 5: you can delete the logical resource entry here as well

templates/llm-router/README.md
```diff
@@ -0,0 +1,1522 @@
...
+!git clone https://github.com/lm-sys/RouteLLM.git /home/ray/default/RouteLLM
+
+# Change to the cloned repository directory
+%cd /home/ray/default/RouteLLM
+
+# Install the package with the specified extras
+!pip install -e .[eval]
+```
+
+    fatal: destination path '/home/ray/default/RouteLLM' already exists and is not an empty directory.
```
akshay-anyscale, line 1007: probably don't need to commit the output cells

anmscale: This one skipped me, good catch! Do you suggest I remove all of them? I kept only a summary showing what the user will see, but maybe not important

akshay-anyscale: yeah I suggest removing all of them unless there's some really important output to display

configs/llm-router/aws.yaml
```diff
@@ -0,0 +1,7 @@
+head_node_type:
```
akshay-anyscale, line 1: you don't need these anymore

akshay-anyscale (approved):

configs/llm-router/aws.yaml
akshay-anyscale: yeah but since you merged into the existing template, you don't need new compute configs at all

anmscale: ok let me remove those files then

scottsun94 (approved):

configs/llm-router/aws.yaml
anmscale: @akshay-anyscale I am not sure what to do about the landing it in the product repo. There I need to specify configs, see e.g. https://github.com/anyscale/product/blob/master/backend/workspace-templates.yaml#L84 and I don't think the configs here: https://github.com/anyscale/product/blob/master/backend/workspace-templates.yaml#L246C12-L246C43 would work

akshay-anyscale: you shouldn't have to make a product repo change for this since the files are in the existing template. Is the only gap that for GCE it doesn't have the serverless config? @kouroshHakha why is that the case?

kouroshHakha: why not make use of basic-server-less configs everywhere? 
```
head_node_type:
  name: head
  instance_type: n1-standard-8
worker_node_types: []
auto_select_worker_config: true
```

kouroshHakha: I don't know why llm finetuning is not on serverless for gce. Maybe that has slipped during transition for some reason? I haven't noticed that until now.

shomilj: I think @anmscale mentioned a GPU head node was a hard req for this workspace; if that has changed, yes, please, let's use serverless :)

anmscale:
Need to rename the branch name to avoid `/` for template to run

PR #206, opened 2024-05-09
Update LoRA learning rates

ArturNiederfahrenhorst:
@SumanthRH flagged to me that our learning rates on the templates are lower. We have been having a too low learning rate on the templates all along.

Thanks @SumanthRH 

Some context on learning rates from instacart workload. TLDR: 1e-4 is a good number.

https://anyscaleteam.slack.com/archives/C035J1Z5EUB/p1711132679445679?thread_ts=1710815945.855269&cid=C035J1Z5EUB

### DIFF
--- a/templates/fine-tune-llm_v2/training_configs/full_param/llama-3-70b.yaml
+++ b/templates/fine-tune-llm_v2/training_configs/full_param/llama-3-70b.yaml
@@ -7,6 +7,7 @@ num_epochs: 10 # <-- change this to the number of epochs that you want to train 
 train_batch_size_per_device: 8
 eval_batch_size_per_device: 8
 learning_rate: 5e-6
+padding: "longest" # This will pad batches to the longest sequence. Use "max_length" when profiling to profile the worst case.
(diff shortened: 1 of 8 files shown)


kouroshHakha:
@ArturNiederfahrenhorst while you are at it :) https://anyscale1.atlassian.net/browse/AE-547 This ticket basically tracked this issue + the lack of documentation of `max_padding`

ArturNiederfahrenhorst:
@kouroshHakha Assuming you mean the line `padding: "max_length"` that we have in other configs: That line does not exist in the templates. It's "longest" by default. Please repeat the precise ask here for me

kouroshHakha:
ok so what I meant is that to include `padding: "longest"` explicitly in all the yamls and then add a comment that says what the other options are, when they should be used. I want people to know that `padding: "max_length"` should be used when doing profiling to make sure they don't oom at the given context length.

kouroshHakha (approved):

ArturNiederfahrenhorst:
I did a test run with this on one of the configs just to be sure, it's fine

PR #214, opened 2024-05-15
Add template for fine-tuning to support function-calling

SumanthRH:
# What does this PR do?

Adds an example for fine-tuning on function calling data through both the Anyscale Platform and through serverless endpoints. I've also added a .gitignore file to the template repo and made some changes to the pre-commit hoooks to support larger notebooks.


kouroshHakha:
A few quick feedbacks: 

- Generally try to match the format of existing templates: https://github.com/anyscale/templates/blob/f939e4664d4b362b7ebdef4754ddd4332e40753c/templates/fine-tune-llm_v2/README.ipynb#L76
- Change title: Function calling fine-tuning on custom data

- Move all the imports to the first section that you want to import. Separate them into two blocks (third party imports and custom module imports)

```
import datasets
import sys
import ray.data 
```

```
from fc_utils.preprocessing import filter_func, initial_mapper, pprint_example, preprocess
```

- In each section in the code block, leave a concise in-block comment on what you are doing.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    INCORRECT_FORMAT = "Incorrect Function Format"
+    INCORRECT_NUMBER_OF_FUNCTION_CALLS = "Incorrect Number of Function Calls"
+    WRONG_FUNCTION_NAME = "Wrong Function Name"
+    WRONG_ARGUMENT_VALUE = "Wrong Argument Value"
+    MISSING_ARGUMENT = "Missing Argument"
+
+    def values():
+        return [item.value for item in PossibleMistakes]
+
+def is_match(response: dict, ground_truth: dict) -> Tuple[bool, str]:
```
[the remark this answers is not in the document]
kouroshHakha: Also for convention let's use the following docstrings format:

```
"""<single line description>

<more description with maybe examples (if really needed)>

Args:
    foo: explain foo line 1
        explain for line 2
    bar: explain bar (note that type is not defined here, it's defined in the signature)

Returns:
    what gets returned + (optional) what gets raised
"""

```

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Checks if the response matches the ground truth. Returns a boolean and a string indicating the mistake type if any.
+    """
+    if ground_truth["tool_calls"] is None:
```
kouroshHakha, line 34: comment each part so that the reader of the code can easily follow, eg:

```
# If ground truth is a normal response, so should be the actual response
```

Also rewrite for better readablity:

```
is_match = True
reason = ""


if x:
    is_match = False
    reason = Foo
elif y:
    is_match = False
    reason = Bar
...

return is_match, reason
```

templates/fine-tune-function-calling/fc_utils/eval_utils.py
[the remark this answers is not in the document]
kouroshHakha: Also the output type is `Tuple[bool, PossibleMistakes]`

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+            return False, PossibleMistakes.WRONG_FUNCTION_NAME.value
+        elif expected_tool_call["arguments"] != actual_tool_call["arguments"]:
+            if len(expected_tool_call["arguments"]) != len(actual_tool_call["arguments"]):
+                return False, PossibleMistakes.MISSING_ARGUMENT.value
+            return False, PossibleMistakes.WRONG_ARGUMENT_VALUE.value
+    return True, ""
+
+
+def parse_and_eval(
+    parser: Union[OpenAIResponseParser, AnyscaleResponseParser],
```
kouroshHakha, line 60: Let's define a base class for these called `ResponseParser` and extend it to build these parsers?

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    messages,
+    expected_responses,
+    tools=None,
+):
+    """
+    Parse and eval loop to parse the assistant responses and evaluate them against the ground truth in the conversation.
+    Args:
+    parser: OpenAIResponseParser or AnyscaleResponseParser object
```
kouroshHakha, line 68: don't forget the indentation
+ first line should be no white-space / new line just a single line concise description.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+            return False, PossibleMistakes.WRONG_ARGUMENT_VALUE.value
+    return True, ""
+
+
+def parse_and_eval(
+    parser: Union[OpenAIResponseParser, AnyscaleResponseParser],
+    messages,
+    expected_responses,
+    tools=None,
+):
```
kouroshHakha, line 64: return type?

.pre-commit-config.yaml
```diff
@@ -4,6 +4,7 @@ repos:
     rev: v3.2.0
     hooks:
       - id: check-added-large-files
+        args: ['--maxkb=10000']
```
kouroshHakha, line 7: what is this for?

templates/fine-tune-function-calling/fc_utils/eval_utils.py
kouroshHakha, line 1: copy the Readme_files to under assets?

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+        return [item.value for item in PossibleMistakes]
+
+def is_match(response: dict, ground_truth: dict) -> Tuple[bool, str]:
+    """
+    Checks if the response matches the ground truth. Returns a boolean and a string indicating the mistake type if any.
+    """
+    if ground_truth["tool_calls"] is None:
+        if response["tool_calls"] is None:
+            return True, ""
+        else:  # explicit else for clarity
```
kouroshHakha, line 37: remove explicit else for clarity?

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+            return False, PossibleMistakes.UNWANTED_FUNCTION_CALL.value
+
+    if response["tool_calls"] is None:
+        return False, PossibleMistakes.NO_FUNCTION_CALL.value
+    elif response["tool_calls"] == INCORRECT_FORMAT: # error during parsing
```
kouroshHakha, line 42: NIT: never use comments that follow code:

Dont do
```
foo = bar # setting foo to bar
```
Instead do
```
# setting foo to bar
foo = bar
```

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+        return False, PossibleMistakes.INCORRECT_FORMAT.value
+    elif len(response["tool_calls"]) != len(ground_truth["tool_calls"]):
+        return False, PossibleMistakes.INCORRECT_NUMBER_OF_FUNCTION_CALLS.value
+
+    for expected_tool_call, actual_tool_call in zip(ground_truth["tool_calls"], response["tool_calls"]):
+        if "name" not in actual_tool_call or "arguments" not in actual_tool_call:
+            return False, PossibleMistakes.INCORRECT_FORMAT.value
+        if expected_tool_call["name"] != actual_tool_call["name"]:
```
kouroshHakha, line 50: elif

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+
+            is_correct, mistake_type = is_match(parsed_response, expected_responses[assist_id])
+            match = match and (is_correct)
+            original_assistant_response = parsed_response["original_response"]
+            current_conv.append(dict(original_assistant_response))
+            assist_id += 1 # next expected response
+            if not match:  # return right away if model output is incorrect
+                return current_conv, match, mistake_type
+        conv_id += 1 # next message
+    return current_conv, match, mistake_type
```
kouroshHakha, line 109: here return is three items but above you are returning two items. Please fix to be consistent.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+                    current_conv[-2]["role"] == "assistant"
+                    and current_conv[-2]["tool_calls"]
+                ):
+                    current_conv[-1]["tool_call_id"] = current_conv[-2]["tool_calls"][0].id
+                else:
+                    return current_conv, False
+            if isinstance(parser, OpenAIResponseParser):
+                parsed_response = parser.get_parsed_response(current_conv, tools)
+            else:
+                parsed_response = parser.get_parsed_response(current_conv)
```
kouroshHakha, line 97: when you define the interface, design it such that it is unified:
```
parsed_response = parser.get_parsed_response(messages, tools)
```
but in the anyscale one don't use tools.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    messages,
+    expected_responses,
+    tools=None,
+):
+    """
+    Parse and eval loop to parse the assistant responses and evaluate them against the ground truth in the conversation.
+    Args:
+    parser: OpenAIResponseParser or AnyscaleResponseParser object
+    messages: list of messages in the conversation. The messages with role 'assistant' are ignored and triggers a chat completion call
+    expected_responses: list of ground truth responses
```
kouroshHakha, line 70: can you add comments / explain why is this a list and not a single assistant message?

kouroshHakha: Also style nit: capitalize the first char.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    tools: list of tools to available, used by OpenAIResponseParser
+    """
+    assert messages[0]["role"] == "system", "First message must be from system"
+    match = True
+    current_conv = []
+    assist_id = 0
+    conv_id = 0
```
kouroshHakha, line 77: nit: for index use idx or index not id. also name them: `cur_message_idx` and `cur_expected_idx`

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+            original_assistant_response = parsed_response["original_response"]
+            current_conv.append(dict(original_assistant_response))
+            assist_id += 1 # next expected response
+            if not match:  # return right away if model output is incorrect
+                return current_conv, match, mistake_type
+        conv_id += 1 # next message
+    return current_conv, match, mistake_type
+
+
+def evaluate_gpt(ds: List[dict], openai_parser: OpenAIResponseParser):
```
kouroshHakha, line 112: use a better name than ds (use compelete names)

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    Evaluates the GPT4 model on the test dataset.
+    """
+    num_correct = 0
+    total_count = 0
+    results = []
+    pbar = tqdm(total=len(ds), desc="Evaluating GPT4..")
```
kouroshHakha, line 119:
```suggestion
    pbar = tqdm(total=len(ds), desc="Evaluating GPT4 ...")
```

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    return results
+
+def evaluate_finetuned(ds: List[dict], anyscale_parser: AnyscaleResponseParser):
+    """
+    Evaluates our fine-tuned model on the test dataset.
+    """
+    anyscale_accuracy = 0
+    total_count = 0
+    results = []
+    pbar = tqdm(total=len(ds), desc="Evaluating Finetuned Model..")
```
kouroshHakha, line 148:
```suggestion
    pbar = tqdm(total=len(ds), desc="Evaluating Finetuned Model ...")
```

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+        if openai_conv is None:  # skip if api errors out
+            openai_is_match = None
+        else:
+            total_count += 1
+        result_dict = {"correct": openai_is_match, "mistake_type": mistake_type, "example": example, "conv": openai_conv}
+        results.append(result_dict)
+
+        if openai_is_match:
+            num_correct += 1
+    print("GPT4 accuracy: ", num_correct / total_count)
```
kouroshHakha, line 138: in jupyter notebooks, don't print stuff in functions and only print them in the notebooks

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,163 @@
...
+    results = []
+    pbar = tqdm(total=len(ds), desc="Evaluating Finetuned Model..")
+    for example in ds:
+        anyscale_messages = example["anyscale_messages"]
+        anyscale_conv, anyscale_is_match, mistake_type = parse_and_eval(anyscale_parser, anyscale_messages, example["expected_responses"])
+        pbar.update(1)
+        if anyscale_conv is None: # skip if api errors out
+            anyscale_is_match = None
+        else:
+            total_count += 1
```
kouroshHakha, line 156: unify the return types. Don't mix up None with empty string, etc. Or at least define the interface more consistently.

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
```diff
@@ -0,0 +1,65 @@
+"""
+Function call json extraction utils.
+Reference: https://gist.github.com/kouroshHakha/6cfbe2bf4aaafc5db733d408044f9902#file-create_test_dataset-py
```
kouroshHakha, line 3: don't mention the random reference in the template :)

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    for i, char in enumerate(jsonl_string):
+        if char == '{':
+            if brace_count == 0:
+                start = i
+            brace_count += 1
+        elif char == '}':
+            brace_count -= 1
+            if brace_count == 0:
+                end = i + 1
+                json_strs.append(jsonl_string[start:end])
```
kouroshHakha, line 34: use regex or sth less adhoc and more systematic?

kouroshHakha: Ask AI to improve your code.

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    jsons = [json.loads(json_str) for json_str in json_strs]
+    return jsons
+
+def get_tool_calls_from_response(assistant_content):
+    assistant_content = assistant_content.strip()  # remove trailing whitespaces
+    escaped_tool_call_tags = [re.escape(tag) for tag in TOOL_CALL_TAGS]
+    if assistant_content.startswith(TOOL_CALL_TAGS[0]):
```
kouroshHakha, line 41: what is TOOL_CALL_TAGS[0]??

kouroshHakha: Never hardcode things like this.

kouroshHakha: Let's do some excercise (it will be a good one)

For each function you have defined, make high level docsring with all types hints and everything. Then feed the docstring and signature into AI and see how well it implemented it (catch its mistakes ofc, but I generally found its style much better than mine it will also give you better ideas on how to make code cleaner and easier to follow, because it was trained to do that). Overtime you'll pick up the right style.

kouroshHakha (commented):
Submitting half of my comments (more to come):

templates/fine-tune-function-calling/fc_utils/eval_utils.py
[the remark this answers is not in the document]
SumanthRH: > never use python without specific type hints (e.g. dict: Dict[str, Any] or Dict[str, np.array] or Dict[str, str] ??).

Will add. 

> Also the output type is Tuple[bool, PossibleMistakes]
Actually the output type I ended up using was a string (i passed the enum value as I can use it directly in plotting), but I can switch to the Enum class.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
SumanthRH: Whoa. Fixed.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
SumanthRH: Done!

.pre-commit-config.yaml
SumanthRH: This is because the current limit can be restrictive for notebooks. My notebook's size is currently 1.3MB. I believe the default limit is around 800KB or so. I made it 10MB for a lot of wiggle room.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
SumanthRH: Hmm these were auto-generated from the readme conversion. Let me fix this.

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
SumanthRH: Is this not systematic? i keep track of open braces and then when all are closed, I extract json from the first open brace. It will work with all sorts of special characters etc. 


Regex was my go-to but I've encountered issues with nested JSONs. The format here is that  you have a string which contains one or more jsons separated by commas. You can have nested jsons separated by commas and regex won't work then (atleast the patterns I tried).  The code here is a bit slower but will be correct.

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
SumanthRH: > Never hardcode things like this. 

"TOOL_CALL_TAGS" are the special indicators used during preprocessing ("[TOOL_CALLS]" and "[/TOOL_CALLS]") . Our post-processing code for the responses has to be pre-processing specific! There's no way to extract assistant response jsons if I don't know what tool call tags I used (assistant response can sometimes have valid json, but not a tool call).  

This is not buried deep within the code btw. It's at the top of [preprocessing.py](https://github.com/anyscale/templates/blob/b84ccda3ba3d22faa8a478c75d40526625ed3158/templates/fine-tune-function-calling/fc_utils/preprocessing.py#L11). This is one code file that has to have a dataset specific implementation. 

WDYT?

SumanthRH: > For each function you have defined, make high level docsring with all types hints and everything. Then feed the docstring and signature into AI and see how well it implemented it (catch its mistakes ofc, but I generally found its style much better than mine it will also give you better ideas on how to make code cleaner and easier to follow, because it was trained to do that). Overtime you'll pick up the right style.

Interesting. Let's see :)

.pre-commit-config.yaml
kouroshHakha: Can you add a concise but informative comment on what this flag does?

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
kouroshHakha: This code is not maintainable. You are adding complexity in a place whose concern is not parsing json. So the preference is to find a one / two liner usining an existing popular library to do the parsing. Say tomorrow this part of the code turns out to be problematic, it will be a lot harder for someone to resolve the issue then. So my question to you is what is the complex jsonl_string that you couldn't use regex for?

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
kouroshHakha: I see let's do something like this:

```
class ToolCallsTag(Enum):
       start = "[TOOL_CALLS]"
       end = "[/TOOL_CALLS]"
```

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+    [95mTool list: [0m[{"name": "calculate_tip", "description": "Calculate the tip amount based on bill total and tip percentage", "parameters": {"type": "object", "properties": {"total": {"type": "number", "description": "The total bill amount"}, "tip_percentage": {"type": "number", "description": "The percentage of tip to be given"}}, "required": ["total", "tip_percentage"]}}]
+    
+
+
+Notice how we made use of special indicators "\[TOOL_CALLS\]" and "\[/TOOL_CALLS\]" to denote a tool call (as supposed a regular JSON output from the model). We'll now further process this conversation format and make it compatible with Anyscale Endpoints. The role "tool" will be converted to the role "user" with a special indicator to highlight that this is a tool response. Further, the tool list will be included in the system prompt with special indicators. The following code block handles the necessary preprocessing.
+
+
+```python
+# complete preprocessing step
+processed_ds = preprocess(ray_ds)
```
kouroshHakha, line 163: Use better names here:

so you'll be doing `anyscale_ds = openai_to_anyscale(ray_ds)`
Let's call the first one `openai_ds = glaive_to_openai(ray_ds)`

So these will be the two public functions. Everything else inside of it should be private function indicated by `_` in front of the name. e.g. `_filter_bad_examples`.

Also why is filtering not applied at when creating the openai dataset? It should be like the following flow:

```
raw_glaive_ds -> openai fmt -> filtered_openai -> anyscale fmt
```

templates/fine-tune-function-calling/fc_utils/preprocessing.py
```diff
@@ -0,0 +1,254 @@
...
+    chat_messages = chat_str_to_messages(example["chat"], tool_to_user=True)
+    messages.extend(chat_messages)
+    if messages[-1]["role"] != "assistant":
+        messages = messages[:-1]  # drop last message if from user
+
+    example["messages"] = messages
+    return example
+
+
+def filter_func(example: Dict[str, Any]) -> bool:
```
kouroshHakha, line 183: returns bool but you are returning int.

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+<p align="center">
+  <img src="./assets/templates.png" alt="Templates">
+</p>
+
+
+
+Follow the instructions to run your fine-tuning job.
+
+## Step 2(b): Fine-tuning through serverless endpoints
+First, obtain your credentials from the [Anyscale platform](https://console.anyscale.com/v2/) and upload the training and validation files in the [fine-tuning tab](https://console.anyscale.com/v2/fine-tuning?fine-tuning-tab=files). Make a note of the file IDs for each. This will be passed to the fine-tuning job.
```
kouroshHakha, line 251: They don't need to upload the files separately from the finetuning job. They can upload data at the time of finetuning submision.

kouroshHakha: so instead of going first under files tab, they can just go to jobs tab and create and upload at the same time.

kouroshHakha: So here you are kinda mixing UI and SDK. 

We can just make it SDK only. You can include code for data upload here as well. And just punt on file upload via UI.

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+Make a note of the final checkpoint after fine-tuning (this should be the last line in the logs). You can now spin up the "Deploy LLMs" template which has all the instructions and required dependencies to serve your finetuned model efficiently. You will find the tutorials on [serving LoRA models](https://github.com/anyscale/templates/blob/main/templates/endpoints_v2/examples/lora/DeployLora.ipynb) (if applicable) and on deploying a [custom model](https://github.com/anyscale/templates/blob/main/templates/endpoints_v2/examples/CustomModels.ipynb) helpful. Once you have set up your fine-tuned model as an Anyscale Service, note down the base URL and API key and place them here.
+
+
+```python
+# to be run only if you finetuned on the Anyscale platform
+ANYSCALE_API_KEY="service-api-key-here"
+ANYSCALE_API_BASE="service-url-here" 
+if not ANYSCALE_API_BASE.endswith("/"):
+    ANYSCALE_API_BASE += "/"
+ANYSCALE_API_BASE += "v1"
```
kouroshHakha, line 295: use fstring

kouroshHakha: just comment on what they should paste (don't use the endswith `/` stuff.) Maybe a screen shot with highlights around the text that is important.

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+```python
+## This is only if you finetuned through serverless endpoints
+ANYSCALE_API_BASE = "https://api.endpoints.anyscale.com/v1"
+ANYSCALE_API_KEY = "esecret_yourKeyHere"
+MODEL_ID = "yourModelIdHere" # make sure to not add a stray slash "/"" at the end!
+```
+
+### (Optional) Try out the model via Playground
+
+You can try out your new model in the Playground: https://console.anyscale.com/v2/playground . In the model dropdown, you should be able to see your finetuned model as shown below
```
kouroshHakha, line 327: ok playground experience is limited to sdk finetuned models only, so make it clear.

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+# Step 4: Evaluation
+
+Let's evaluate our trained model with GPT-4 as a baseline. 
+
+
+## Evaluation strategy
+
+Evaluation of function calling capability is non-trivial, given that we're looking to extract structured data from an inherently unpredictable and unstructured stream of text. We will use the following simple evaluation strategy: The models are evaluated on the accuracy metric and their responses are graded as accurate if their response for each assistant entry in the conversation is correct. An assistant response is graded as correct under the below conditions:
+1. In case the ground truth response contains no function call, then the model's response should also not have a function call. 
+2. In case the ground truth response contains a function call, then the model's response should also have a function call. We do not check for the content of the response here. The assistant function call should further have the correct function name and the correct function arguments.
```
kouroshHakha, line 342: > We do not check for the content of the response here.
This is confusing? I know you mean the "content" in the message but instead of this just say we only check the name and arguments of the function call to ensure correctness.

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+# enter your openai key below.
+OPENAI_API_KEY = "yourApiKeyHere" 
+# enter your Anyscale key below. If you finetuned through Anyscale endpoints, you can get the key here: https://console.anyscale.com/credentials. Otherwise, you should use the key from your Anyscale Service
+ANYSCALE_API_KEY = "yourApiKeyHere" 
+```
+
+
+```python
+# initialize parsers
+openai_parser = OpenAIResponseParser(api_key=OPENAI_API_KEY, api_base="https://api.openai.com/v1", model="gpt-4", tool_call_tags=TOOL_CALL_TAGS)
```
kouroshHakha, line 391: "https://api.openai.com/v1" use a const definition above where you define the ANYSCALE_API_BASE?

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+anyscale_parser = AnyscaleResponseParser(api_key=ANYSCALE_API_KEY, api_base=ANYSCALE_API_BASE, model=MODEL_ID, tool_call_tags=TOOL_CALL_TAGS) 
+```
+
+
+```python
+# evaluate both models and plot the results
+results_gpt, accuracy_gpt = evaluate_model(modified_ds, openai_parser, "gpt")
```
kouroshHakha, line 398: separate teh evaluate_model in different blocks, because this takes time.

kouroshHakha (commented):
Notebook feedback:
I have left reviews on the README.md. Please make the changes to the notebook such that the generated README will get fixed. 

Code architecture feedback:
- Let's revise the data flow: 
The initial mapper currently goes from glaive dataset to an arbitray format that is neither openai neither anyscale. What I would do? I would create an API that looks like:  ray_ds.map(glaive_to_openai) and implement glaive_to_openai to be the dataset format compatible with openAI. Right now your assistant content includes [TOOL_CALL] tags which is anyscale specific while you also have a tool role in your messages. The issue is that you are combining structured with unstructured together here, which means you have to carry the complexity around all over the place. You can avoid it by being smarter at the beginning.

Let's also make it clear (with a graph) that we are doing: 

glaive -> openAI -> Anyscale

in terms of formatting. 

- I see that you wrote the pprint_example in a way that accept keys (probably because the example that you pass into the pprint includes other things you don't want to print). The reason for this is that your initial_mapper is not implemented properly. Why are you keeping the old keys in the return dict from initial_mapper?

Style feedback:

- Make sure your IDE wraps lines at 80 characters. Then go through each file and see if wrapping has happened. If so please break the lines

You'll find other specific comments inside the code.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,207 @@
...
+        - example: The original example from the dataset
+        - conv: The full conversation with the model's actual response.
+
+    Args:
+        dataset: List of examples to evaluate
+        openai_parser: ResponseParser object
+
+    Returns:
+        results: List of results
+    """
```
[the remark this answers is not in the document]
kouroshHakha: Note that in docstring above, first line is a short (less than 60-70 character) description.

templates/fine-tune-function-calling/fc_utils/preprocessing.py
```diff
@@ -0,0 +1,254 @@
...
+    if not keys:
+        keys = example.keys()
+    pprint_str = ""
+    # ANSI escape code for blue, green, red, yellow and magenta colors
+    blue = "\033[94m"
+    reset = "\033[0m"
+    green = "\033[92m"
+    red = "\033[91m"
+    magenta = "\033[95m"
+    yellow = "\033[93m"
```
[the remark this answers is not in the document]
SumanthRH: done! have also taken care of different data formats now so that this looks much nicer.

templates/fine-tune-function-calling/fc_utils/preprocessing.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    Preprocesses the Ray dataset into the messages format required by Anyscale Endpoints.
+    """
+    ray_ds = ray_ds.map(final_mapper)
+    ray_ds = ray_ds.filter(filter_func)
+    ray_ds = ray_ds.drop_columns(["system", "chat"])  # drop the original columns
+    return ray_ds
+
+
+def pprint_example(example: Dict[str, Any],keys: List[str] = []) -> None:
```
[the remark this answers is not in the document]
SumanthRH: Hmm yes this was not meant to be purely openai. The flow I had in my mind was to follow the dataset's format and then change it to something like the openai format and then bring to anyscale. It's now changed to exactly the openai format and then our format.

templates/fine-tune-function-calling/fc_utils/preprocessing.py
```diff
@@ -0,0 +1,254 @@
...
+            if tool_to_user:
+                # add tool tags only if the tool response is to be converted to user role
+                function_response = f"{TOOL_RESULT_TAGS[0]} {function_response} {TOOL_RESULT_TAGS[1]}"
+                role = "user"
+            msg = {"role": role, "content": function_response}
+        messages.append(msg)
+    return messages
+
+
+def final_mapper(example: Dict[str, Any]) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
SumanthRH: Yes this is actually better. I took the path of least resistance first and the intermediate dataset doesn't add much information. Have changed `initial_mapper` to `glaive_to_openai` now.

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
```diff
@@ -0,0 +1,95 @@
...
+class FunctionCallNotFoundError(Exception):
+    pass
+
+class FunctionResponseNotFoundError(Exception):
+    pass
+
+def extract_jsons(string: str) -> List[Dict[str, Any]]:
+    """
+    Extracts JSON objects from a string containing one or more JSONs.
+    Example: "[{"name": "weather", "arguments": {\"location\": \"New York\"}}]"
```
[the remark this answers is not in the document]
SumanthRH: Hmm the problem I had was that I wanted to use this function in both places: for the glaive system message and for extracting tool calls from assistant responses. In the case of the former, the jsons are separated by a space and in the latter, I wanted to have them comma separated (more standard). This made things complicated with regex. I've now stuck to using regex for the system message through `extract_functions_from_system_msg` and then made two sets of parsers for assistant calls in glaive and openai/anyscale responses.

.pre-commit-config.yaml
SumanthRH: Done!

templates/fine-tune-function-calling/fc_utils/preprocessing.py
```diff
@@ -0,0 +1,254 @@
...
+            msg = {"role": "user", "content": user_content}
+        elif match.group("assistant"):
+            assistant_content = match.group("assistant").strip()
+            assistant_content = combine_multiple_entries(assistant_content)
+            # glaive dataset is full of single function calls.
+            # We convert the single function call into a list and add tool call tags
+            if GLAIVEAI_TEMPLATE["tool_call_prefix"] in assistant_content:
+                # make function call a list and add tags
+                assistant_content = assistant_content.replace(
+                    GLAIVEAI_TEMPLATE["tool_call_prefix"], f"{TOOL_CALL_TAGS[0]} ["
```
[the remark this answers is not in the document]
SumanthRH: Yes I actually wrote that this is an Openai-like conversation format.  The flow I had in my mind was to stay close to the dataset's format and then change it to something _like_ the openai format and then bring to anyscale. It's now changed to exactly the openai format and then small modifications for our format.

templates/fine-tune-function-calling/fc_utils/plot_utils.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+import numpy as np
+import matplotlib.pyplot as plt
+
+from fc_utils.eval_utils import Mistakes
+
+ # 10 colors for different flags
+COLORS = ['salmon', 'deepskyblue', 'springgreen', 'orange', "gold", "limegreen", "turquoise", "blueviolet", "darkred", "darkorange"]
+
+def plot_results(results_finetuned, results_gpt):
```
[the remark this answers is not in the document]
SumanthRH: On it! Will standardize a result format as well. This plotting function was initially a second-class member of the utilities package, so I'll try to make this neater.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,207 @@
...
+    if model not in ["gpt", "finetuned"]:
+        raise ValueError(f"Model must be either 'gpt' or 'finetuned', got {model}")
+    num_correct = 0
+    total_count = 0
```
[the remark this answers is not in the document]
SumanthRH: Hmm I wanted to avoid wasting memory, but agreed that the above approach is more pythonic and looks cleaner. Have switched to this.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,207 @@
...
+        # query the model, parse and evaluate the generated responses
+        conv, is_match, mistake_type = parse_and_eval(
+            parser, messages, example["expected_responses"], tools=tools
+        )
+
+        # skip if api errors out
+        if conv is None:
+            invalid_entry = True
+        else:
+            # valid, increment total count
```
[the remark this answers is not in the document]
SumanthRH: Taking care of this now!

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,419 @@
...
+<p align="center">
+  <img src="./assets/distr_glaive_pie.png" alt="Distribution">
+</p>
+
+
+# Table of Contents
+1. [Data Preprocessing](#step-1-data-preprocessing)
+2. [Finetuning](#step-2-fine-tuning)
+3. [Serving](#step-3-serving)
+4. [Evaluation](#step-4-evaluation)
```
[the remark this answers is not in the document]
SumanthRH: In terms of the flow, this doesn't make much sense for me. Since we want the user to evaluate the model they fine-tuned in Step 2, serving should come before evaluation (otherwise what endpoint will `ANYSCALE_API_BASE` be)

templates/fine-tune-function-calling/README.md
SumanthRH: Ah... I actually missed this part of the SDK! Have updated the upload instructions now.

templates/fine-tune-function-calling/README.md
SumanthRH: Good idea. I just kept it to be a subsection for the serverless endpoints finetuning, but added an explicit note.

templates/fine-tune-function-calling/README.md
SumanthRH: Have added this along with the API keys section above.

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
SumanthRH: A part of the problem was that I was using the same function in some places to process Glaive's system message (where the jsons are space separated) and our formatted tool calls (where they are comma separated). I've refactored the whole thing now to have use to separate the two kinds of formats. I've also used one common function to process our tool info formatting (which is "[prefix_content] <tag1> [tool info] <tag2>") for tool lists, tool calls, tool results, etc. and then leave further parsing to `json`.

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
SumanthRH: I've put a higher-level NamedTuple called IndicatorTags and used that to instantiate our specific tool list, tool call and tool result tags.

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,207 @@
...
+    INCORRECT_NUMBER_OF_FUNCTION_CALLS = "Incorrect Number of Function Calls"
+    WRONG_FUNCTION_NAME = "Wrong Function Name"
+    WRONG_ARGUMENT_VALUE = "Wrong Argument Value"
+    MISSING_ARGUMENT = "Missing Argument"
+    NONE = ""
+
+    def values():
+        return [item.value for item in Mistakes]
+
+def check_match(response: Dict[str, Union[str, ToolCallType]], ground_truth: Dict[str, Union[str, ToolCallType]]) -> Tuple[bool, Mistakes]:
```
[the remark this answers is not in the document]
SumanthRH: Done!

templates/fine-tune-function-calling/fc_utils/eval_utils.py
```diff
@@ -0,0 +1,207 @@
...
+                        reason = Mistakes.WRONG_ARGUMENT_VALUE
+    return is_match, reason
+
+
+def check_format(tool_calls: List[ToolCallType]) -> bool:
+    """
+    Checks if the tool call is in the correct format.
+    """
+    for tool_call in tool_calls:
+        if "name" not in tool_call or "arguments" not in tool_call or not isinstance(tool_call["arguments"], dict) or not isinstance(tool_call["name"], str):
```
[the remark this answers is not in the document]
SumanthRH: Done! Since this function also enforces a specific tool format, I've moved it to my data formatting/constants util file `data_format.py`

templates/fine-tune-function-calling/fc_utils/function_extraction_utils.py
```diff
@@ -0,0 +1,95 @@
...
+Function call json extraction utils.
+"""
+
+import json
+import re
+from typing import List, Dict, Any, Tuple, TypeAlias, Union
+
+
+# define our custom type for a tool call
+ToolCallType: TypeAlias = Dict[str, Union[str, Dict[str, str]]]
```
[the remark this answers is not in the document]
SumanthRH: Good catch! Removed the type hint

templates/fine-tune-function-calling/fc_utils/preprocessing.py
```diff
@@ -0,0 +1,325 @@
...
+    tools = example["tools"]
+    # Convert the stringified tools to a list of jsons
+    tools = json.loads(tools) if tools else []
+    for message in openai_messages:
+        if message["role"] == "system":
+            anyscale_message = {"role": "system", "content": message["content"]}
+            if len(tools):
+                tools_str = json.dumps(tools)
+                tools_str_with_tags = (
+                    f"{TOOL_LIST_TAGS[0]} {tools_str} {TOOL_LIST_TAGS[1]}"
```
kouroshHakha, line 191: TOOL_LIST_TAGS.start and end?

templates/fine-tune-function-calling/fc_utils/preprocessing.py
```diff
@@ -0,0 +1,325 @@
...
+import json
+from typing import Dict, Any, List, Optional
+import ray.data
+from fc_utils.function_extraction_utils import (
+    get_tool_calls_from_response,
+    extract_functions_from_system_msg,
+    FunctionCallNotFoundError,
+    DatasetFormat,
+)
+from fc_utils.data_format import *
```
kouroshHakha, line 15: never do `from x import *` always be explicit about imports.

templates/fine-tune-function-calling/README.md
```diff
@@ -0,0 +1,478 @@
...
+
+If you notice, each sample has two entries: system and chat. This dataset is already formatted in specific way (e.g. using USER, \<|endoftext|\> and other tokens). To enable fine-tuning on various open source models we need to convert each row to a more general format like the OpenAI chat format, which is the preferred format for fine-tuning instruction-tuned models on Anyscale ([dataset format guide](https://docs.endpoints.anyscale.com/endpoints/fine-tuning/dataset-prep)). Let's first bring this dataset into the conversation format and inspect how that looks like:
+
+
+```python
+import fc_utils.preprocessing
+import importlib
+importlib.reload(fc_utils.preprocessing)
+from fc_utils.preprocessing import pprint_example
+```
```
kouroshHakha, line 129: what are these? can be removed?

kouroshHakha (commented):
Really good. A few comments left.

templates/fine-tune-function-calling/fc_utils/preprocessing.py
SumanthRH: Hmm that rule can have exceptions.  Especially with a case like here where one file has mostly constants and classes all used in the other file (preprocessing.py). I've switched to the explicit import anyways (though it's ugly and unnecessary!).

templates/fine-tune-function-calling/fc_utils/test_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    return {"messages": processed_messages}
+
+
+def get_evaluation_dataset(
+    test_ds: ray.data.Dataset,
+    tool_call_tags: IndicatorTags,
+    tool_result_tags: IndicatorTags,
+    tool_list_tags: IndicatorTags,
+    format: DatasetFormat,
+):
```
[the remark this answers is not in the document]
SumanthRH: Outdated! Was in the middle of pushing these fixes

templates/fine-tune-function-calling/README.md
SumanthRH: I was still cleaning this up! Removed now with all new changes incorporated.

templates/fine-tune-function-calling/fc_utils/preprocessing.py
SumanthRH: Yes done. I was still cleaning up!

templates/fine-tune-function-calling/README.md
SumanthRH: Have fixed the wording now.

templates/fine-tune-function-calling/README.md
SumanthRH: Yup added.

kouroshHakha:
I'd like add one thing on top of Goku's comments:

> I'm not sure how unavoidable this since the dataset is quite involved is but it was a bit difficult to follow the evaluation workload: evaluate_model --> parse_and_eval --> get_parsed_response --> get_completion --> check_match. Is there a way to consolidate this?

Can we create a flow-chart in the readme that shows the process so that the reader can understand the architecture of the design if they want to dive in the code.

templates/fine-tune-function-calling/README.ipynb
```diff
@@ -0,0 +1,761 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Step 3: Serving\n",
+    "\n",
+    "## Step 3(a): Finetuned on the Anyscale Platform\n",
+    "\n",
+    "Make a note of the final checkpoint after fine-tuning (this should be the last line in the logs). You can now spin up the \"Deploy LLMs\" template which has all the instructions and required dependencies to serve your finetuned model efficiently. You will find the tutorials on [serving LoRA models](https://github.com/anyscale/templates/blob/main/templates/endpoints_v2/examples/lora/DeployLora.ipynb) (if applicable) and on deploying a [custom model](https://github.com/anyscale/templates/blob/main/templates/endpoints_v2/examples/CustomModels.ipynb) helpful. Once you have set up your fine-tuned model as an Anyscale Service, head over to the \"Services\" tab in the console and select your deployed service. \n",
```
akshay-anyscale, line 435: Can you update this based on the new flow in the template with a generate config script?

SumanthRH: Hmm this won't be a standalone template so just I'm directing them to the serve template - that will have all the instructions. So wouldn't it be better to keep this as is?

SumanthRH:
@GokuMohandas @kouroshHakha Addressing your comments here:

> I'm not sure how unavoidable this since the dataset is quite involved is but it was a bit difficult to follow the evaluation workload: evaluate_model --> parse_and_eval --> get_parsed_response --> get_completion --> check_match. Is there a way to consolidate this?

> Can we create a flow-chart in the readme that shows the process so that the reader can understand the architecture of the design if they want to dive in the code.

I hope the eval flowchart clarifies why we have the current setup!

> Also in the error analysis, would it worth showing show a base model would perform? I can appreciate that the fine-tuned model and GPT-4 both perform roughly the same number of mistakes, but I'd love to see how bad the original base model was.

Good idea, I've added this! The base model system prompt is not completely optimized, but initial results show a big gap:

image

I've also added some small explainers for how I'm evaluating the base model, let me know if it makes sense (Also, you can take a look at the system prompt here: https://github.com/anyscale/templates/blob/df456a0ccc4ce32bc016c841dae90fda5e52992c/templates/fine-tune-function-calling/fc_utils/data_format.py#L18 

> Also when we serve these models, do you have to apply the chat template manually or at least ensure that theprompt_format is correct in our model_config_<timestamp> file? I may have missed this if you do it somewhere already.


We're not touching this at all! Data processing is operating at the OpenAI-chat interface so I'm only figuring out how the "user" "system" responses (and "assistant" responses in case of training) look like. Applying the chat template is left to our internal fine-tuning and serving code (which is consistent).

templates/fine-tune-function-calling/README.ipynb
kouroshHakha: +1 to @SumanthRH

kouroshHakha:
I have not spelled check things again. The plots / figures add much better narration impact on the reader. So approved.

akshay-anyscale (approved):

PR #224, opened 2024-05-24
[DO NOT MERGE] Move to using the new llm-forge fine-tuning API

ArturNiederfahrenhorst:
llmforge has a new API "llmforge finetune ...." that encapsulates the bits of hackery we had in the main.py file in the template.

Prerequisite: https://github.com/anyscale/llm-forge/pull/353/files


ArturNiederfahrenhorst:
I've updated this PR to respect the next llm-forge release, which is still to happen.

PR #226, opened 2024-05-28
Add Docs team to CODEOWNERS for templates/

angelinalg:
Add Docs team to codeowners for templates/

PR #230, opened 2024-05-31
Revert "Add Docs team to CODEOWNERS for templates/"

angelinalg:
Reverts anyscale/templates#226

Going to postpone being a blocker until we have more cycles. We are at capacity with finishing up the docs migration. We will do this again later, and with the correct handle.


Aydin-ab:

PR #233, opened 2024-06-04
Updated template model names

shawnanyscale:
- Updated template model names


kouroshHakha (commented):
Can you update v1 as well?

shawnanyscale:
> Can you update v1 as well?

v1 actually doesn't have configs for llama3. https://github.com/anyscale/templates/tree/main/templates/fine-tune-llm/training_configs

kouroshHakha (approved):

PR #238, opened 2024-06-07
Install fixed version of `optimum` in `text-embeddings` template

scottjlee:
In order to use the default `ray-ml` image in the workspace, we need to also install `optimum`, which is used in the template.


omatthew98 (approved):
🚀

PR #249, opened 2024-06-14
[Triton] Add performance benchmark for triton vs pytorch

GeneDer:
Added performance benchmark for triton

Closes https://github.com/ray-project/ray/issues/45716


templates/triton_services/pytorch_app.py
```diff
@@ -0,0 +1,56 @@
...
+            use_safetensors=True,
+        ).to("cuda")
+        self.pipe.unet.to(memory_format=torch.channels_last)
+
+        if run_compile:
+            self.pipe.unet = torch.compile(
+                self.pipe.unet, mode="reduce-overhead", fullgraph=True,
+            )
+            # This will compile the model to reduce overhead and improve performance.
+            _ = self.pipe("test").images[0]
```
edoakes, line 41: I believe torch.compile can add overhead for multiple calls. would suggest that your benchmark throws out the first N requests for each condition

templates/triton_services/pytorch_app.py
```diff
@@ -0,0 +1,56 @@
...
+from ray import serve
+from diffusers import DiffusionPipeline
+from fastapi.responses import StreamingResponse
+
+app = FastAPI()
+
+
+@serve.deployment(
+    ray_actor_options={
+        "num_gpus": 1,
```
edoakes, line 22: not needed; default

templates/triton_services/pytorch_app.py
```diff
@@ -0,0 +1,56 @@
...
+    @app.get("/generate")
+    async def generate(self, prompt: str) -> StreamingResponse:
+        print(f"Generating image for prompt: {prompt}")
+        image_ = self.pipe(prompt).images[0]
+
+        # Stream back the image to the caller.
+        buffer = BytesIO()
+        image_.save(buffer, "JPEG")
+        buffer.seek(0)
+        return StreamingResponse(buffer, media_type="image/jpeg")
```
edoakes, line 52: hmm.. why use streaming response if you seem to be generating the full image on line 46?

templates/triton_services/pytorch_app.py
```diff
@@ -0,0 +1,56 @@
...
+            self.pipe.unet = torch.compile(
+                self.pipe.unet, mode="reduce-overhead", fullgraph=True,
+            )
+            # This will compile the model to reduce overhead and improve performance.
+            _ = self.pipe("test").images[0]
+
+    @app.get("/generate")
+    async def generate(self, prompt: str) -> StreamingResponse:
+        print(f"Generating image for prompt: {prompt}")
+        image_ = self.pipe(prompt).images[0]
```
edoakes, line 46: this should be run in an executor. you can either do it explicitly or just make the function `def` instead of `async def`

templates/triton_services/pytorch_app.py
GeneDer: Yea, from my test the very first call is super slow (took ~ 1 minute) That's why I added this to the constructor so the any real call from the benchmark are fast 😄

templates/triton_services/pytorch_app.py
GeneDer: Bc we are sending bytes back. I actually haven't test if that image can be returned directly. Let me try it out

edoakes: yeah just return https://www.starlette.io/responses/#plaintextresponse with media_type set

templates/triton_services/pytorch_app.py
GeneDer: Just double check this is talking about "num_cpus" being default and not needed right? "num_gpus" is still needed just so the deployment can be scheduled onto GPU nodes.

edoakes: yes

GeneDer:
@edoakes this is ready for another look when you have some time🙏

templates/triton_services/triton_app.py
```diff
@@ -51,7 +50,7 @@ def generate(self, prompt: str) -> StreamingResponse:
             buffer = io.BytesIO()
             image_.save(buffer, "JPEG")
```
edoakes, line 51: why the `_` suffix on image?

templates/triton_services/triton_app.py
```diff
@@ -51,7 +50,7 @@ def generate(self, prompt: str) -> StreamingResponse:
...
             buffer.seek(0)
```
edoakes, line 52: don't think this is needed if you're calling `getvalue()`

templates/triton_services/triton_app.py
```diff
@@ -51,7 +50,7 @@ def generate(self, prompt: str) -> StreamingResponse:
             buffer = io.BytesIO()
```
edoakes, line 50: converting to an `Image` type and then writing it into the buffer seems pointless here -- you already have a raw byte array above, why not just return that directly? does `fromarray` do any actual transformation of the data?

edoakes (approved):
image data munging looks odd to me otherwise lgtm

templates/triton_services/triton_app.py
GeneDer: it's a copypasta, let me took it out

templates/triton_services/triton_app.py
edoakes: what does it do exactly? numpy array of uint8 is already a byte buffer

GeneDer: Talked offline, this seems to be official/ easiest way to converting RGB array into JPEG image bytes 🙂

PR #250, opened 2024-06-20
[ft-template] Add long-context data prep cookbook

ArturNiederfahrenhorst:
This PR adds a cookbook that showcases to users what to do if their dataset contains many examples that are "long".
This should not appeal to users who usually end up with context-length of 512 but ones, but ones on the opposite end of the spectrum who ask themselves if their dataset will fit.


ArturNiederfahrenhorst:
Adding a video walkthrough

templates/fine-tune-llm_v2/cookbooks/long_context/README.ipynb
```diff
@@ -0,0 +1,115 @@
...
+    "**⏱️ Time to complete**: 90 minutes (assuming GCP A2 nodes can be acquired immidiately)\n",
+    "\n",
+    "This guide demonstrates how to prepare a dataset with long context lengths.\n",
+    "Make sure you have gone over the [basic fine-tuning guide](../../README.md) before going over this cookbook.\n",
+    "\n",
+    "In the following example, we filter a hugginface dataset and assemble it into the right format.\n",
+    "By estimating how many tokens an example from the dataset will result in, we can disgard examples that are too long.\n",
+    "You can use this as a template for creating your own datasets.\n",
+    "\n",
+    "> **_NOTE:_** To fine-tune with a context length of 8k tokens and the llama-3-8b.yaml config in this cookbook, we require GCP A2 nodes. For that, you need to instantiate your workspace in a region such as GCP's `us-central1`  due to the relatively high availability. See [GCP's availability info](https://cloud.google.com/compute/docs/gpus/gpu-regions-zones) for more information. If you only want to generate the dataset, you can do that anywhere."
```
kouroshHakha, line 17: Again here, just mention the resource requirement. How they should get A100s (GCP or not) should be asked in a different thread and that's beyond the scope of this cookbook.

templates/fine-tune-llm_v2/cookbooks/long_context/README.ipynb
```diff
@@ -0,0 +1,115 @@
...
+   "outputs": [],
+   "source": [
+    "import json\n",
+    "import pandas as pd\n",
+    "from transformers import AutoTokenizer\n",
+    "\n",
+    "# Fill in your personal hugginface token with access to the tokenizer (You can use a similar tokenizer as a work-around)\n",
+    "HHUGGINFACE_TOKEN = \"\"\n",
+    "# The name of the model you want to fine-tune with. We use this only for tokenization so models with the same tokenizer are interoperable here.\n",
+    "MODEL_NAME = \"meta-llama/Meta-Llama-3-8B-Instruct\"\n",
```
kouroshHakha, line 33: Llama-3-8B already is 8k context. So what is the point ?

templates/fine-tune-llm_v2/cookbooks/long_context/README.ipynb
```diff
@@ -0,0 +1,115 @@
...
+    "\n",
+    "### How do I find out how many tokens the examples in my dataset have?\n",
+    "\n",
+    "If you fine-tune with this template, you will find exact statistics at the beginning of your fine-tuning job.\n",
+    "Exact numbers are hard to compute in advance. To get a rough idea, you can instantiate a tokenizer for your model and feed it a few samples from your dataset. Remember that one example consists of system message, a prompt, and an answer. There may also be [online services](https://belladoreai.github.io/llama3-tokenizer-js/example-demo/build/) that can help.\n",
+    "\n",
+    "### What if my dataset results in examples that are longer that the native context length of the model?\n",
+    "\n",
+    "Some datasets have examples that, when tokenize, exceed the context length of your LLM.\n",
+    "Anyscale Endpoints and the Anyscale platform support extending the native context length of LLMs. For example, for Llama 3 8B, we support fine-tuning with up to 32768 tokens (see [Anyscale Endpoints](https://docs.anyscale.com/canary/endpoints/fine-tuning/supported-models/) docs for a list).\n",
```
kouroshHakha, line 84: Anyscale endpoints is gone. We should just completely remove any new references to that.

kouroshHakha (commented):
I just read this cookbook and I am lost about what the exact focus is. 

Is it context extension? If so then you should focus on the functionality and detailed abstraction on how it is done in the backend (like using Rope scaling etc). All user needs to do is to make sure their dataset is properly formatted at longer context and provide the desired context length in their configs. 

There is no need to go through the dataset details, just give an example already prepared dataset that has this length. 

Show the statistics (average token length, etc) of the dataset and make sure that the importance of this cookbooks clicks with the user.

templates/fine-tune-llm_v2/cookbooks/long_context/README.ipynb
ArturNiederfahrenhorst: Are you absolutely sure about this? If we remove this, users will have to find out themselves which cloud will give them a good chance of getting A100s. If we don't but it here at the start, they will create their workspace in AWS only to find out later that they need to create it again in GCP because you can't just move workspaces around.

templates/fine-tune-llm_v2/cookbooks/long_context/README.ipynb
```diff
@@ -0,0 +1,115 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Fine-tuning on datasets with long context\n",
```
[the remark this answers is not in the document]
ArturNiederfahrenhorst: This cookbook is not about extending the context though and I have avoided that wording everywhere afaics. You mention in your review:
> I just read this cookbook and I am lost about what the exact focus is.
>  Is it context length extension?
It's not context-length extension or as your comment suggests here ("extending context length"). I only mention extention in the FAQ because of that. We have always handled context-length extension as something that should be transparent to the user. We can tell them that quality suffers the more they exceed the native context length and that we use rope-scaling, I agree with that! But that's literally one sentence.

The focus is how to prepare your dataset if you want to fine-tune with context lengths that are so long that users worry about them. When you say "All user needs to do is to make sure their dataset is properly formatted at longer context" - I think that's THE part in a users journey where they might need some help. 

If I understand you correctly, we want to scrap this part, where we show how to chip away parts of your dataset that are too long. Instead, we should demonstrate choosing a dataset that has a context length that exceeds the native context length of a mode. With that in mind, I think we can scrap this cookbook alltogether and just make this a one-liner the main readme.

templates/fine-tune-llm_v2/cookbooks/long_context/README.ipynb
ArturNiederfahrenhorst: The point is that, to users, context extension should be transparent.
Just like it was on endpoints. There is no point in using a 16k dataset if it's transparent to them anyway.

The point of this notebook is to show users how to prepare their dataset if examples are rather long.
8k context length is already quite the long context length and we are still talking about a template here. That is, this is also meant to be an examples that people can just run. It takes about twice as much time to grab the necessary amount of A100s to fine-tune at 16k. And it will also take much longer to fine-tune. That's why I chose 8k after all.

ArturNiederfahrenhorst:
Closing because I'm not on this team anymore and context has likely been lost

PR #251, opened 2024-06-21
Fix `wandb` API key export in finetuning template

SumanthRH:
# What does this PR do?

This is a tiny PR to fix `wandb` key export in the finetuning template. Currently, we exported the key through an iPython shell command `!export WANDB_API_KEY=xxx` . This, however, does not actually accomplish what we want, because training is launched with a separate shell command `!python main.py ...` . Each shell command gets executed in a different subshell, which means `WANDB_API_KEY` does not transfer over to the training script env. This PR fixes the issue by exporting env var through `os`.

### DIFF
--- a/templates/fine-tune-llm_v2/README.ipynb
+++ b/templates/fine-tune-llm_v2/README.ipynb
@@ -47,7 +47,8 @@
    "outputs": [],
    "source": [
     "# [Optional] You can set the WandB API key to track model performance\n",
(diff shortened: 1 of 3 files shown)


kouroshHakha (commented):
There are other places we have used this pattern? Can you do a global search and fix all of them?

SumanthRH:
Looks like I missed the same command in our cookbook! Added this now. Other templates don't have this issue.

kouroshHakha (approved):

PR #252, opened 2024-06-24
Update batch sizes for llama 3 8B to counter pytorch allocator cache flushes

ArturNiederfahrenhorst:
While releasing the 0.5.0.0 llm-forge image, we realized that we can some "_pytorch allocator cache flushes_" warnings.
According to pytorch, these slow down training and should be eliminated.
In some instances, they have also led to GPU OOMs.

Process behind this PR:
I launched every possible configuration (lora and full-parameter).
If the warning occured at the beginning, I'd let the training continue for a couple of iterations to see if we stabilize and the warning goes away. If it goes away, I didn't take action (in accordance with the pytorch warning message).
If the warning would occur from time to time, I'd reduce batch size.

I went through this process for all models and configurations to dogfood the template and see if this problem occurs elsewhere. The problem only occured for llama 3 8b (lora and full-param).


kouroshHakha (approved):

SumanthRH (approved):
LGTM. Im assuming the new batch sizes still maximize GPU util.

PR #253, opened 2024-06-27
[Triton] drop gpu header dependency

GeneDer:
Wrap model compile and upload in a ray actor to run on the GPU worker so we can remove the GPU head node dependency on this template. Also fix the configs. 

Full walkthrough video: https://drive.google.com/file/d/1HGOFNYneniECHuP4iX43_IsjiW9zYZRj/view?usp=sharing

### DIFF
--- a/configs/triton_services/aws.yaml
+++ b/configs/triton_services/aws.yaml
@@ -3,16 +3,5 @@ head_node_type:
   instance_type: m5.2xlarge
   resources:
     cpu: 0
-# TODO: revert this
-#worker_node_types: []
-#auto_select_worker_config: true
-worker_node_types:
-  - name: worker-node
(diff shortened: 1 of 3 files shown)


shomilj (commented):
can you just delete configs/triton_services/aws.yaml and replace with default serverless_config

shomilj (commented):
thanks for doing this!

GeneDer:
> can you just delete configs/triton_services/aws.yaml and replace with default serverless_config

Good call, made the change in product repo to use the serverless configs https://github.com/anyscale/product/pull/29368/commits/82f181483492cf3f43fbd362a475805199fca65a

shomilj (approved):
sweet!

PR #274, opened 2024-07-20
added ray data batch classification template 

soffer-anyscale:
Added the Ray Data Batch Inference template as part of the "Run Quickstart" project.

This project will test out linking the popular examples on the Ray Docs to hosted Anyscale templates so users can click a "Run Quickstart" button that will open up the template on hosted Anyscale.

This is also part of the OSS Ray/K8s -> Hosted Anyscale Funnel project.

This PR includes only one template so I can test out adding the templates. Shortly, we'd want at least the top 10 public examples to follow.


configs/ray-data-batch-classification/aws.yaml
```diff
@@ -0,0 +1,8 @@
+head_node_type:
+  name: head
+  instance_type: g5.xlarge
+worker_node_types:
+- name: gpu_worker
+  instance_type: g5.xlarge
+  min_workers: 1
+  max_workers: 3
```
aslonnie, line 8: add an eol at the end of this file

configs/ray-data-batch-classification/gcp.yaml
```diff
@@ -0,0 +1,8 @@
+head_node_type:
+  name: head
+  instance_type: n1-standard-4-nvidia-t4-16gb-1
+worker_node_types:
+- name: gpu_worker
+  instance_type: n1-standard-4-nvidia-t4-16gb-1
+  min_workers: 1
+  max_workers: 3
```
aslonnie, line 8: and also here

.gitignore
```diff
@@ -157,7 +157,7 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+.idea/
```
aslonnie, line 160: why change this?

configs/ray-data-batch-classification/aws.yaml
```diff
@@ -0,0 +1,8 @@
+head_node_type:
+  name: head
+  instance_type: g5.xlarge
```
shomilj, line 3: Can this use the serverless auto select workers mode? We are trying to push away from GPU head nodes due to idle cost, especially when using these for community edition type use cases.

.gitignore
soffer-anyscale: I was using PyCharm and it creates the .idea folder, we can keep it commented if needed.

configs/ray-data-batch-classification/aws.yaml
soffer-anyscale: Yes, definitely.

soffer-anyscale (commented):
Updated files based on feedback.

configs/ray-data-batch-classification/aws.yaml
aslonnie: is this serverless now? seems that it is still using gpu head node.

aslonnie: I think what @shomilj meant is to not use a gpu node as head node, and run workloads on worker nodes with auto gpu node selection.

aslonnie (commented):
removing myself from review list. you need a template codeowner to review this.

scottsun94 (approved):

PR #277, opened 2024-07-26
grammar: "runs within a workspace"

aslonnie:
rather than "is run within a workspace"

### DIFF
--- a/templates/batch-llm/README.ipynb
+++ b/templates/batch-llm/README.ipynb
@@ -16,7 +16,7 @@
     "\n",
     "For a Python script version of the code in this workspace template, refer to `main.py`.\n",
     "\n",
-    "**Note:** This tutorial is run within a workspace. Review the `Introduction to Workspaces` template before this tutorial.\n",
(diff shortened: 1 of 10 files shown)


aslonnie:
or maybe it should be "This tutorial is **a** run within a workspace" ? I feel that using run in verb form is more direct.

scottsun94:
Will leave this to @emmyscode or @angelinalg

emmyscode (changes requested):
@angelinalg Can we actually take this as an opportunity to remove the reference to an example running in a workspace, so that it doesn't read weirdly when it's just in docs?

angelinalg:
Thanks for filing this! We'll add it to the backlog for sure.

aslonnie:
> take this as an opportunity to remove the reference to an example running in a workspace, so that it doesn't read weirdly when it's just in docs?

any specific action item for this PR?

emmyscode (approved):
Let's just get this PR in for now. No blocker on this.

angelinalg:
No action.

aslonnie:
ok, let me merge this, and I will try to go through the template release steps.

PR #280, opened 2024-08-01
[RayLLM] Added tool calling examples based on JSON Mode

kouroshHakha:
This PR adds an end-to-end example to the templates for demonstrating function calling. 

With this example in, we don't need to support function calling as a end-to-end feature on RayLLM anymore. This is more transparent and more developer friendly. All they need is JSON mode to be enabled from the engine. They can further hack it together to get better results.


kouroshHakha:
The content of this PR has been merged directly into the main through another PR. Closing this one.

PR #282, opened 2024-08-05
Merge from main

akshay-anyscale:


configs/ray-data-batch-classification/aws.yaml
```diff
@@ -0,0 +1,5 @@
+head_node_type:
+  name: head
+  instance_type: m5.2xlarge
+worker_node_types: []
+auto_select_worker_config: true
```
shomilj, line 5: not sure when this got merged, but this should be just using the default serverless config instead of defining a new config - @scottsun94

shomilj (approved):
codew

kouroshHakha:
This should wait on an rayllm release. @shrekris-anyscale will create a separate branch to make sure the sequencing is respected for rayllm.

shrekris-anyscale (approved):

PR #288, opened 2024-08-12
[`batch-llm`] Check for private/gated models during HuggingFace token authentication

scottjlee:
Some models on HuggingFace are private/gated models. Previously, the HuggingFace API returned a 400-status code in the case where the model is private/gated and no user token is provided, but the API now returns a 200 status code, with the response now including `private` and `gated`. 

This PR updates the utility function used in the `batch-llm` template to determine if the model requires a HF token so that it works correctly for private/gated models with the new HF API response.


scottjlee:
Closed in favor of https://github.com/anyscale/templates/pull/289

PR #300, opened 2024-08-19
[llm-ft] make the cross refs of examples point to ipynb instead of readme for doc rendering reasons

kouroshHakha:


templates/fine-tune-llm_v2/README.ipynb
```diff
@@ -57,8 +57,8 @@
     "\n",
     "Here is a list of end-to-end examples that involve more steps such as data preprocessing, evaluation, etc but with a main focus on improving model quality via fine-tuning.\n",
     "\n",
-    "* [Fine-tuning for Function calling on custom data](./end-to-end-examples/fine-tune-function-calling/README.md): This example demonstrates how to fine-tune a model on a custom dataset for function calling task.\n",
-    "* [Preference tuning for summarization](./end-to-end-examples/fine-tune-preference/README.md): An example of Direct Preference Optimization (DPO) fine-tuning for summarization."
+    "* [Fine-tuning for Function calling on custom data](./end-to-end-examples/fine-tune-function-calling/README.ipynb): This example demonstrates how to fine-tune a model on a custom dataset for function calling task.\n",
```
SumanthRH, line 60: My understanding was that the doc deployment is erroring out because of the relative links in the md file. For example, an error in a recent PR of mine: https://github.com/anyscale/docs/actions/runs/10461229527/job/28969099563#step:7:176 

This is still a relative link so won't it fail again?

kouroshHakha: There are two issues. we diverged from the contract of matching a pattern of `"(examples/.*.ipynb)"`. I am relaxing the folder part to include `end-to-end-examples` but the file extension should remain .ipynb (I tried md but that broke something else)

SumanthRH: oh so it detects these and automatically gets full link? sgtm!

SumanthRH (approved):

PR #312, opened 2024-09-03
Keep readme markdown and notebook in sync.

aslonnie:
use relative asset links

### DIFF
--- a/templates/intro-jobs/README.md
+++ b/templates/intro-jobs/README.md
@@ -59,7 +59,7 @@ You can view active and historical job runs at (`Home > Jobs`). Click into the j
 
 You should see the job state and its output on the overview page.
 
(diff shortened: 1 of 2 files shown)


emmyscode (commented):
How do we prevent the images from breaking in the console (when the images are edited on main, but still the same relative path in release branch/tag) and docs?

aslonnie:
> How do we prevent the images from breaking in the console (when the images are edited on main, but still the same relative path in release branch/tag) and docs?

not sure if I understand the question correctly, but:

- console are not referencing the main branch today?
- asset urls are being replaced with loading the readme file
- I do have a plan to embed all the images inline so that it won't worry about the `assets/` directory.

emmyscode:
Never mind on the console concern. For docs, since we sync the templates READMEs in, we need to change this script https://github.com/anyscale/docs/blob/master/sync_oa_templates.py to switch from relative links for assets into absolute so that we can display images.

PR #313, opened 2024-09-03
[E2E LLM] Update template with `anyscale.llm.dataset` and `anyscale.llm.model`

galenhwang:
* upward dependency on https://github.com/anyscale/.../pull/30646
* see https://docs.google.com/document/d/1RhyKa1C9A98M5exdh-5J0Ysl_aHC7tvGYeVDQD22mHc/edit for discussion

---

Updates template with below SDK calls:
- `anyscale.llm.dataset.upload()`
- `anyscale.llm.dataset.download()`
- `anyscale.llm.model.get()`

before, we were loading datasets from `ANYSCALE_ARTIFACT_STORAGE`, which was only a local variable. datasets are now a first-class API citizen and can be retrieved with `anyscale.llm.dataset` API

a later PR will make further improvements to the template:
- improve the template flow
  - eg. removing the need to copy/paste values (eg. `model_id`) from logs
  - substitute `train_dataset` and `valid_dataset` in the yaml automatically

**manual test**
confirmed on staging with latest `anyscale` CLI changes
(will need to rebuild the E2E image with the latest `anyscale` package)


templates/e2e-llm-workflows/README.ipynb
```diff
@@ -512,12 +491,52 @@
...
+    "]:\n",
+    "    dataset.write_json(temp_path, filename_provider=CustomFilenameProvider(filename))\n",
+    "    def get_full_name():\n",
+    "        full_name = os.environ.get(\"ANYSCALE_USERNAME\", \"default\")\n",
+    "        full_name = \"_\".join(full_name.split())  # Join the full name with underscores\n",
+    "        full_name = full_name.lower()  # Lowercase the full name\n",
+    "        return full_name\n",
+    "    \n",
+    "    # full_name/e2e_llm/viggo/train.jsonl\n",
+    "    dataset_name = f\"{get_full_name()}/e2e_llm/viggo/{filename}\"\n",
```
akshay-anyscale, line 525: I'd set the name to something more user-friendly like viggo-training instead of a file path

templates/e2e-llm-workflows/README.ipynb
```diff
@@ -512,12 +491,52 @@
...
+    "            \"VIGGO dataset for E2E LLM template, \"\n",
+    "            f\"'{filename}' split\"\n",
+    "        ),\n",
+    "    )\n",
+    "    config_fields[field_name] = dataset.storage_uri\n",
+    "    rprint(f\"Metadata for '{filename}'\")\n",
+    "    rprint(dataset)\n",
+    "rprint(\"Config fields:\")\n",
+    "rprint(config_fields)"
    ]
```
akshay-anyscale, line 540: let's also update the "Load artifacts" section with the models sdk

templates/e2e-llm-workflows/README.ipynb
```diff
@@ -917,7 +941,8 @@
    ],
    "source": [
     "# Load test set for eval\n",
-    "ft_test_ds = ray.data.read_json(f\"{os.environ['ANYSCALE_ARTIFACT_STORAGE']}/viggo/test.jsonl\")\n",
+    "test_path = config_fields[\"test_path\"]\n",
+    "ft_test_ds = ray.data.read_json(test_path)\n",
```
akshay-anyscale, line 945: shouldn't this use datasets instead of the config fields?

templates/e2e-llm-workflows/README.ipynb
```diff
@@ -917,7 +941,8 @@
...
     "test_data = ft_test_ds.take_all()\n",
     "test_data[0]\n"
    ]
```
akshay-anyscale, line 948: another thing to demonstrate the use of datasets is to write eval results back to a dataset that can be used after to calculate metrics

templates/e2e-llm-workflows/README.ipynb
galenhwang: this would be fine for ray summit training day

but for regular use case, it would use the default project of the cloud they're in, so users in the same cloud could be overriding (uploading new versions) of this `viggo-training` if they're working with the template at the same time

templates/e2e-llm-workflows/README.ipynb
galenhwang: sg, @SumanthRH is tracking this

templates/e2e-llm-workflows/README.ipynb
galenhwang: yeah, that's a good callout

i'll improve it to save the `anyscale.llm.dataset.Dataset` object from `anyscale.llm.dataset.upload()` to a var called `test_dataset`

```py
test_dataset = anyscale.llm.dataset.upload(...)
...
ft_test_ds = ray.data.read_json(test_dataset.storage_uri)
```

templates/e2e-llm-workflows/README.ipynb
akshay-anyscale: maybe we can call it viggo-training-{user}

galenhwang: sure, that seems reasonable
```py
    dataset_name = f"viggo-training-{get_full_name()}",
```

keep in mind - they can also modify the dataset name to something of their own choice

templates/e2e-llm-workflows/README.ipynb
galenhwang: will address in separate PR, see [PR descr.](https://github.com/anyscale/templates/pull/313#issue-2503507658)

> - improve the template flow
>   - eg. removing the need to copy/paste values (eg. model_id) from logs

akshay-anyscale (approved):

templates/e2e-llm-workflows/README.ipynb
galenhwang: i see, i'm probably overthinking it then, i'll do `viggo/{filename}`

i think a path-like structure is fine b/c our models are named `meta.../llama...`, so they're used to this structure already

templates/e2e-llm-workflows/README.md
```diff
@@ -54,10 +54,13 @@ os.environ['HF_TOKEN'] = ''  # <-- replace with your token
 ray.init(runtime_env={'env_vars': {'HF_TOKEN': os.environ['HF_TOKEN']}})
 ```
 
-```bash
-(autoscaler +15m27s) [autoscaler] Downscaling node i-0267f387d8cf66c2e (node IP: 10.0.21.202) due to node idle termination.
-(autoscaler +15m28s) [autoscaler] Cluster resized to {8 CPU, 0 GPU}.
-```
+    2024-06-12 16:52:33,877	INFO worker.py:1564 -- Connecting to existing Ray cluster at address: 10.0.46.70:6379...
```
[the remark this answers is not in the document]
galenhwang: these logs are auto-printed on `ray.init()`

also we add pre-existing logs so users have some expectation on what the cell will output when run

templates/e2e-llm-workflows/README.md
```diff
@@ -446,26 +496,27 @@ from src.utils import download_files_from_bucket
 ```python
 # Locations
 artifacts_dir = '/mnt/cluster_storage'  # storage accessible by head and worker nodes
-model = 'meta-llama/Meta-Llama-3-8B-Instruct'
-uuid = 'gokum:ueewk'  # REPLACE with your NAME + MODEL ID (from Job logs)
-artifacts_path = (
-    f"{os.environ['ANYSCALE_ARTIFACT_STORAGE'].split(os.environ['ANYSCALE_CLOUD_STORAGE_BUCKET'])[-1][1:]}"
-    f"/lora_fine_tuning/{model}:{uuid}")
+model_id = model_info.id 
+artifacts_path = f"{os.environ['ANYSCALE_ARTIFACT_STORAGE']}/lora_fine_tuning/{model_id}"
```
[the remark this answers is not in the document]
SumanthRH: Actually no! The models API doesn't know about our checkpoint forwarding. So the `storage_uri` in `model_info` looks like :

```
s3://anyscale-production-data-cld-ldm5ez4edlp7yh4yiakp2u294w/org_4snvy99zwbmh4gbtk64jfqggmj/cld_ldm5ez4edlp7yh4yiakp2u294w/artifact_storage/eric__tang/llmforge-finetuning/meta-llama/Meta-Llama-3-8B-Instruct/TorchTrainer_2024-09-10_18-12-20/epoch-0
```

This happens because we treat checkpoint forwarding and the model registry insert as two independent callbacks so they don't know about each other.  Will need to be fixed on the llm-forge side, but this is not bad for now I think?

templates/e2e-llm-workflows/README.md
```diff
@@ -749,16 +765,18 @@ We'll start by generating the configuration for our service. We provide a conven
...
 
-**LoRA weights storage URI**: `s3://anyscale-test-data-cld-i2w99rzq8b6lbjkke9y94vi5/org_7c1Kalm9WcX2bNIjW53GUT/cld_kvedZWag2qA8i5BjxUevf5i7/artifact_storage/lora_fine_tuning`
 
-**model**: `meta-llama/Meta-Llama-3-8B-Instruct:gokum:atyhk`
+We can use the model metadata `model_info` for the model ID as well the storage URI for the final lora weights.  
+
+**model**: `meta-llama/Meta-Llama-3-8B-Instruct:suman:cdrba` (`model_info.id`)
 
+**LoRA weights storage URI**: `s3://org_4snvy99zwbmh4gbtk64jfqggmj/cld_ldm5ez4edlp7yh4yiakp2u294w/artifact_storage/lora_fine_tuning/meta-llama/Meta-Llama-3-8B-Instruct:suman:cdrba` ( or `{ANYSCALE_ARTIFACT_STORAGE}/lora_fine_tuning/{model_id}`)
```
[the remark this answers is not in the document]
SumanthRH: Ah I see the issue. "the storage uri" term has been used to refer to the full path where the current checkpoint is stored. but here we're talking about rayllm dynamic lora loading path. 

Let me clarify the wording and fix the path

templates/e2e-llm-workflows/README.md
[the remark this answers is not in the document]
akshay-anyscale: hmm this is definitely something we should fix. My assumption was the checkpoint upload to s3 is just part of the model registry insert (which is what you'd expect if you were using the sdk from outside llmforge)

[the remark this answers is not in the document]
akshay-anyscale: how do you envision the model registry works if you don't use llmforge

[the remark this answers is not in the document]
SumanthRH: > My assumption was the checkpoint upload to s3 is just part of the model registry insert  (which is what you'd expect if you were using the sdk from outside llmforge)

Hmmm good point. Yeah right now these are distinct operations that do not know about each other. The lora checkpoint forwarding has been treated as a rayllm detail separate from the model registry insert. But this can indeed be done at model insert time. 

> (which is what you'd expect if you were using the sdk from outside llmforge)

I agree that when they interact with the SDK, they'd want the model metadata to be self contained for use in inference. This is not the case right now. 

> how do you envision the model registry works if you don't use llmforge

Hmm thinking about this, I'm wondering if the forwarding can indeed be done in the product backend before the model insertion in the DB? That way this API is self contained in registering a model for use with rayllm. Otherwise llmforge will need to encode this information and you end up with corner cases like this.

templates/e2e-llm-workflows/README.md
```diff
@@ -406,38 +406,88 @@ max_retries: 0
...
-```
 
-```bash
-Output
-(anyscale +0.8s) Submitting job with config JobConfig(name='llm-fine-tuning-guide', image_uri='localhost:5555/anyscale/llm-forge:0.4.3.2', compute_config=None, env_vars=None, py_modules=None).
-(anyscale +3.2s) Uploading local dir '.' to cloud storage.
-(anyscale +4.8s) Job 'llm-fine-tuning-guide' submitted, ID: 'prodjob_515se1nqf8ski7scytd52vx65e'.
-(anyscale +4.8s) View the job in the UI: https://console.anyscale.com/jobs/prodjob_515se1nqf8ski7scytd52vx65e
-(anyscale +4.8s) Use `--wait` to wait for the job to run and stream logs.
+!anyscale job submit --config-file "deploy/jobs/ft.yaml" --exclude "assets"
```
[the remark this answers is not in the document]
galenhwang: modified to SDK call in https://github.com/anyscale/templates/pull/313/commits/d87083f703e3316dd69bba3493f9d966b5dd4dd2

templates/e2e-llm-workflows/README.ipynb
galenhwang: fixed in https://github.com/anyscale/templates/pull/313/commits/d87083f703e3316dd69bba3493f9d966b5dd4dd2

templates/e2e-llm-workflows/README.ipynb
```diff
@@ -778,28 +1062,90 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "036ec442",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import anyscale\n",
+    "model_info = anyscale.llm.model.get(job_id=\"prodjob_5i8jpkxa6pyit9mzzzzry31iae\")\n",
```
galenhwang, line 1083: @SumanthRH you can do `job_id=job_id` now - it's initialized in an above cell

SumanthRH: Done!

templates/e2e-llm-workflows/README.md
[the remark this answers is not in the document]
SumanthRH: Done!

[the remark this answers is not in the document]
SumanthRH: Tracking here: https://anyscale1.atlassian.net/browse/AE-1128 

For the current template, I'm gonna retain the existing code.

templates/e2e-llm-workflows/README.md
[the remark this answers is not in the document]
SumanthRH: Tracking here: https://anyscale1.atlassian.net/browse/AE-1128 

For the current template, I'm gonna retain the existing code.

templates/e2e-llm-workflows/README.md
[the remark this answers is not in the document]
SumanthRH: Done!

templates/e2e-llm-workflows/README.ipynb
SumanthRH: Done!

PR #317, opened 2024-09-12
Add logging with MLflow and example configs for WandB/MLflow to finetuning template

erictang000:
Adding separate section in README for logging (since it's not on by default). Removed wandb from first command to have separate logging section/instructions more clear. Maybe need to link to docs/api reference once those are up

[Example WandB run from template](https://wandb.ai/anyscale-llm-forge/llmforge/runs/btv1k2cz?nw=nwusereatang)
[Example MLflow run from template](https://dbc-073b287d-29d2.cloud.databricks.com/ml/experiments/3140617323020126/runs/177cd940312b4281acca3d2d38f77bd0?o=2200977071475194)


templates/fine-tune-llm_v2/README.ipynb
```diff
@@ -33,7 +33,36 @@
...
+    "```bash\n",
+    "pip install mlflow\n",
+    "```\n",
+    "\n",
+    "Then we can set the `MLFLOW_TRACKING_TOKEN` environment variable instead of `WANDB_API_KEY`, and specify `mlflow` as the `logger` in the config. We've provided an example config at `training_configs/custom/meta-llama/Meta-Llama-3-8B/lora/4xA10-512-mlflow.yaml` showing how to additionally configure the `logger` field with a valid `tracking_uri` and `experiment_name`.\n",
+    "\n",
+    "Then we can launch finetuning as above with our modified config:\n",
+    "```bash\n",
+    "MLFLOW_TRACKING_TOKEN=<PUT_YOUR_MLFLOW_KEY_HERE> llmforge anyscale finetune training_configs/custom/meta-llama/Meta-Llama-3-8B/lora/4xA10-512-mlflow.yaml\n",
+    "```"
```
SumanthRH, line 65: I feel like the intro page here shouldn't start having these feature specific sections. The current way we've organized this is to 
- First document the feature in our intro page in the [docs](https://docs.anyscale.com/llms/finetuning/intro/). 
- Add example configs here if needed
- Add a user guide on the docs which will optionally reference configs here. 
- Make sure there's some visibility to this feature in the template (optional)

One example: https://docs.anyscale.com/llms/finetuning/guides/bring_any_hf_model 


Suggestion:
- This PR will add example configs and just mention the user guide for logging (this note is special for logging, given that there's been a breaking change)
- Update the intro page to highlight new mlflow support 
-  In the user guide, let's update it now and reference the example configs as 

 _We've provided an example config for wandb in the [fine-tuning template](url). To try out fine-tuning with logging on wandb, run
`llmforge anyscale finetune <path_to_config>`_ 


Unfortunately these changes will need to wait until patch release to fix the logger bug (wil push to make it happen soon), but we can keep the material ready.

erictang000: ok so here should I just add the WandB logger to the default config (4xA10-512.yaml), and then we can just add a note mentioning MLflow and linking to the docs

SumanthRH: Hmm actually the default path can include no logger as well, I think that's fine.

SumanthRH: But we need to add a note on logger for sure, but it's just going to be linking to the user guide on the docs

templates/fine-tune-llm_v2/training_configs/custom/meta-llama/Meta-Llama-3-8B/lora/4xA10-512-mlflow.yaml
```diff
@@ -0,0 +1,67 @@
...
+# Deepspeed configuration, you can provide your own deepspeed setup
+deepspeed:
+  config_path: deepspeed_configs/zero_3_offload_optim+param.json
+
+logger:
+  provider: mlflow
+  provider_config:
+    tracking_uri: "https://*******.cloud.databricks.com"
+    experiment_name: "/Users/*******/llmforge"
+    create_experiment_if_not_exists: True
```
SumanthRH, line 42: Hmm in general, we've always made configs that are runnable by default withexample datasets, etc. I guesss this is unavoidable for MlFLow given that the tracking uri is internal to their MLFlow account? 

In any case we should document that this needs to change

```suggestion
# Logging config
logger:
  provider: mlflow
  provider_config:
     # Make sure to change this to your tracking uri! 
    tracking_uri: "https://*******.cloud.databricks.com"
    # Make sure to change this to the path where experiments should be logged!
    experiment_name: "/Users/*******/llmforge"
    create_experiment_if_not_exists: True
```

SumanthRH (commented):
Left a comment.

SumanthRH (approved):

PR #318, opened 2024-09-16
[ci] pin python and package versions

aslonnie:
so that they do not easily break due to Internet changes


aslonnie:
I wonder if we should migrate the CI to buildkite just for unity.

emmyscode (approved):

PR #320, opened 2024-09-17
[maketmpl] add CI setup for new templates building tool

aslonnie:
we are using golang for the following reasons:

- separate it from python dependencies in this repository, which are
  used by the templates
- it is easier to build a single binary and run it against other
  branches in the past; properly packaging a python program with its
  dependencies is much harder without introducing additional tools
- it is just faster


.github/workflows/premerge.yaml
```diff
@@ -0,0 +1,49 @@
+name: premerge
```
aslonnie, line 1: this file is renamed from `pre-commit.yaml`

aslonnie:
the full draft PR is here:
https://github.com/anyscale/templates/pull/315

this is just to set up the skeleton.

PR #321, opened 2024-09-18
[tmpls] standardize img tag format

aslonnie:
so that downstream post-processing scripts can process the README.md files and transform the img tags properly.


templates/batch-llm/README.ipynb
```diff
@@ -423,7 +423,7 @@
     "### Monitoring Dataset execution\n",
     "We can use the Ray Dashboard to monitor the Dataset execution. In the Ray Dashboard tab, navigate to the Job page and open the \"Ray Data Overview\" section. Click on the link for the running job, and open the \"Ray Data Overview\" section to view the details of the batch inference execution:\n",
     "\n",
-    "<img src=\"assets/ray-data-jobs.png\" width=900px/>"
+    "<img src=\"assets/ray-data-jobs.png\" width=900px />"
```
aslonnie, line 426: based on html syntax guide, there needs to be a space between `px` and `/>`, otherwise, it will be parsed into `width="900px/"` (which is how python or go's html parsing will return when parsing the string..)

although, web browsers are often super tolerant on these kind of non-standard writing..

templates/stable-diffusion-pretraining/01_preprocess.ipynb
```diff
@@ -6,7 +6,7 @@
    "source": [
     "# Scalable data preprocessing pipeline for Stable Diffusion\n",
     "\n",
-    "<img src=\"https://anyscale-materials.s3.us-west-2.amazonaws.com/stable-diffusion/preprocessing_architecture_v4.jpeg\" width=\"900px\">\n",
+    "<img src=\"https://anyscale-materials.s3.us-west-2.amazonaws.com/stable-diffusion/preprocessing_architecture_v4.jpeg\" width=900px />\n",
```
aslonnie, line 9: we have some downstream scripts that assume that these width and heights are not quoted... :( the downstream script is the one doing the wrong thing (in theory), but it is easier to fix the issue here upstream.

https://github.com/anyscale/docs/blob/master/sync_oa_templates.py#L127

emmyscode (approved):
Looks good!

vigneshka (approved):
lgtm

PR #322, opened 2024-09-18
[maketmpl] parsing images in markdown files.

aslonnie:
for semi-reliably parsing and generating embedded img from
readme files


aslonnie:
cc @can-anyscale

khluu:
I don't know much about HTML so it's better to leave this for someone who knows more frontend to review!

ci/maketmpl/md_image.go
```diff
@@ -0,0 +1,76 @@
...
+
+	src string
+	alt string
+
+	heightPx string
+	widthPx  string
+
+	style string
+
+	isHTML bool
```
lanbochen-anyscale, line 23: seems `isHTML` always true? can be removed?

aslonnie: it is used later. non-html / md image tags are directly parsable with regex, so it is much simpler.. but let me add it in here. give me a minute.

aslonnie: added the part that is parsing images from entire md file now.

aslonnie: and now the field is used :)

lanbochen-anyscale (approved):

PR #327, opened 2024-09-18
Ray Summit 2024: Stable Diffusion pre-training

marwan116:


configs/ray-summit-stable-diffusion/aws.yaml
```diff
@@ -0,0 +1,17 @@
...
+    CPU: 16
+    GPU: 1
+  flags: {}
+worker_nodes:
+  - instance_type: g5.4xlarge
+    flags: {}
+    name: '1xA10G:16CPU-64GB'
+    min_nodes: 1
+    max_nodes: 1
+    market_type: ON_DEMAND
```
shomilj, line 14: Curious why you don't want to use serverless / resource-driven scaling for this?

Also, isn't hard-coding the `cloud` here going to cause issues?

configs/ray-summit-stable-diffusion/aws.yaml
```diff
@@ -0,0 +1,17 @@
...
+enable_cross_zone_scaling: false
+flags:
+  allow-cross-zone-autoscaling: false
```
shomilj, line 17: Only the `enable_cross_zone_scaling` needs to be present here, and I would set it to `true`

marwan116: This is the output of "export yaml" 

So for context, we intend to perform an on-demand capacity reservation in a specific AZ and we want to ensure we make use of these resources. 

Are you suggesting we keep it to true, just in case as a backup if capacity is reached ?

configs/ray-summit-stable-diffusion/aws.yaml
marwan116: @shomilj please let me know if this resolved 🙏

configs/ray-summit-stable-diffusion/aws.yaml
marwan116: @shomilj given the context I provided about ODCR do you still recommend setting `enable_cross_zone_scaling` to true ?

configs/ray-summit-stable-diffusion/aws.yaml
shomilj: Actually, I think this entire schema is not right -- and this won't work when you try running it end to end. Downloading the YAML from the UI and pasting it here is not workable - the YAML export follows the convention of the new SDK, and this workspace template stuff bypasses the SDK and gets sent straight to a backend representation. So you want to probably copy the structure from https://github.com/anyscale/templates/blob/main/configs/e2e-llm-workflows/aws.yaml.

And also before merging -- you should test this end to end somehow. I'm not sure how, maybe check with workspaces team.

configs/ray-summit-stable-diffusion/aws.yaml
shomilj: Ok if you have OCDR then you can probably disable. But you may need to either ask users to hard-code a zone...since I don't think you can fix the zone in this compute config, so the SIM will pick any zone (not necessarily the one with your OCDR).

configs/ray-summit-stable-diffusion/aws.yaml
```diff
@@ -0,0 +1,14 @@
+head_node_type:
+  instance_type: g5.4xlarge
+  resources:
+    cpu: 12
+    gpu: 1
```
shomilj, line 5:
```suggestion
```

shomilj (approved):

PR #328, opened 2024-09-18
Ray Summit 2024: End-to-End LLMs

marwan116:


shomilj (approved):
assuming you want to intentionally disable head node scheduling on this one

PR #336, opened 2024-09-20
[maketmpl] add utility function to build a zip file

aslonnie:
from a list of files. will use this to build template release artifacts

### DIFF
--- a/ci/maketmpl/util.go
+++ b/ci/maketmpl/util.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 )
 
 func checkIsDir(path string) error {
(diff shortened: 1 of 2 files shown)


khluu (approved):

ci/maketmpl/util.go
```diff
@@ -34,3 +35,25 @@ func addFileToZip(z *zip.Writer, file, pathInZip string) error {
 	}
 	return nil
 }
+
+func buildZip(dir string, files []string, out string) error {
```
thomasdesr, line 39: Thoughts for future: might have been slightly nicer to use `os.DirFS` & `zipfile.Writer.AddFS()` for this?

aslonnie: well, I actually need to ignore some files.. so I will need to write custom fs's anyways.. feel that it is more flexible, direct and straightforward this way.

PR #351, opened 2024-09-25
[maketmpl] add readme notebook parsing and writing

aslonnie:
a python notebook can be converted into two different flavor's of markdown files, one is a github flavored, for checking in and updating in the source, when is for releasing, with all images inlined, to be used in anyscale product and docs pages


ci/maketmpl/readme.go
```diff
@@ -0,0 +1,141 @@
...
+
+	md   []byte
+	imgs []*mdImage
+}
+
+func (f *readmeFile) writeInto(w io.Writer, imgOpts *writeImgOptions) error {
+	cursor := 0
+	for i, img := range f.imgs {
+		if cursor < img.start {
+			if _, err := w.Write(f.md[cursor:img.start]); err != nil {
```
lanbochen-anyscale, line 23: add comment what this mean `f.md[cursor:img.start]`?

lanbochen-anyscale (approved):

ci/maketmpl/readme.go
aslonnie: added comments that explains this function.

this function is copying the markdown with all the images rewritten.

PR #357, opened 2024-09-26
[ci] refactor ci bash script

aslonnie:
no conversion result change; just better bash scripting.

### DIFF
--- a/ci/auto-generate-readme.sh
+++ b/ci/auto-generate-readme.sh
@@ -1,45 +1,63 @@
 #!/bin/bash
+
+set -euo pipefail
+
 echo "Auto-generating README files..."
 
-# Search for notebook files named README.ipynb in the ../templates directory
-notebook_files=$(find ../templates -name "README.ipynb")
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+if [[ "$(pwd)" != "${REPO_ROOT}" ]]; then
+	echo "Must run this script at repo's root directory".
+	exit 1
+fi
 
-# Loop through each notebook file
-for notebook_file in $notebook_files; do
-    # Exclude specific notebooks from conversion
-    if [ "$notebook_file" != "../templates/templates/getting-started/README.ipynb" ] && [ "$notebook_file" != "../templates/templates/e2e-llm-workflows/README.ipynb" ] && ! grep -q "Time to complete" $notebook_file; then
-        echo "**********"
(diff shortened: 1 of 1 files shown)


aslonnie:
need to refactor this file so that I can add fake readme in `maketmpl` testdata for unit testing..

ci/auto-generate-readme.sh
```diff
@@ -1,45 +1,63 @@
...
-    fi
-    if [ "$notebook_file" != "../templates/templates/e2e-llm-workflows/README.ipynb" ]; then
-        # Convert notebook file to README.md using nbconvert
-        jupyter nbconvert --to markdown "$notebook_file" --output-dir "$(dirname "$notebook_file")"
-    else
-        echo "Skipping README generation for $notebook_file"
-    fi
-done
+# Search for notebook files named README.ipynb in the ../templates directory
+TEMPLATES_DIRS=($(find "templates" -mindepth 1 -maxdepth 1 -type d))
```
khluu, line 14: just wondering why do we need another pair of parentheses here?

khluu, line 14: just wondering why do we need another pair of parentheses here?

ci/auto-generate-readme.sh
```diff
@@ -1,45 +1,63 @@
...
+	if [[ ! -f "${TMPL}/README.ipynb" ]]; then
+		echo "README.ipynb file not found; skipping notebook conversion and checking."
+	else
+        # Exclude specific notebooks from conversion
+        TMPL_NAME="$(basename "${TMPL}")"
+        NOTEBOOK_FILE="${TMPL}/README.ipynb"
+
+        if [[ "${TMPL_NAME}" == "getting-started" || "${TMPL_NAME}" == "e2e-llm-workflows" || "${TMPL_NAME}" == "ray-summit-multi-modal-search" ]]; then
+            echo "Skip 'Time to complete' checking for ${TMPL_NAME}"
+        elif ! grep -q "Time to complete" "${NOTEBOOK_FILE}" ; then
```
khluu, line 32: maybe double bracket here

khluu (approved):
LGTM syntax wise!

ci/auto-generate-readme.sh
aslonnie: the inner one is capturing the output, the outer one converts it into a bash array.

khluu: oh got it

PR #360, opened 2024-09-26
Remove some friction from e2e LLM template

RCdeWit:
- Remove potentially breaking `pip install -U anyscale`, since it's no longer needed.
- Automatically parse in the S3 bucket locations to the train config yaml

### DIFF
--- a/templates/e2e-llm-workflows/src/utils.py
+++ b/templates/e2e-llm-workflows/src/utils.py
@@ -5,6 +5,7 @@ from tempfile import TemporaryDirectory
 import boto3
 from urllib.parse import urlparse
 from ray.data import Dataset
+import yaml
 
 
 def download_files_from_s3(s3_uri, local_dir):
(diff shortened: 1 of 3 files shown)


SumanthRH (approved):
Thanks Rob!

templates/e2e-llm-workflows/src/utils.py
```diff
@@ -63,3 +64,20 @@ def get_dataset_file_path(dataset: Dataset):
...
         yield dataset_file_path
+
+def set_key_value_in_config_file(config_path: str, key: str, value: str):
+    """Take a path to a .json config file and update the key to a value, then save the file"""
+    with open(config_path, 'r') as stream:
+        try:
+            loaded = yaml.safe_load(stream)
+        except yaml.YAMLError as exc:
+            print(exc)
+            exit()
```
SumanthRH, line 75: Since we're not handling the exception, how about just leaving it as is for the user to handle? 
```suggestion
    with open(config_path, 'r') as stream:
        loaded = yaml.safe_load(stream)
```

templates/e2e-llm-workflows/src/utils.py
```diff
@@ -63,3 +64,20 @@ def get_dataset_file_path(dataset: Dataset):
...
+    # Modify the fields from the dict
+    loaded[key] = value
+    # Save it again
+    with open(config_path, 'w') as stream:
+        try:
+            yaml.dump(loaded, stream, default_flow_style=False)
+        except yaml.YAMLError as exc:
+            print(exc)
```
SumanthRH, line 83: Same comment here: 
```suggestion
    with open(config_path, 'w') as stream:
        yaml.dump(loaded, stream, default_flow_style=False)
```

PR #361, opened 2024-09-26
[maketmpl] add support to save a different file or inlined bytes in zip

aslonnie:
makes it possible to put a properly converted readme file into the built zip archive


aslonnie:
cc @vigneshka

lanbochen-anyscale (approved):

PR #381, opened 2024-10-10
Add DSPy e2e template

isaacbmiller:
Most relevant file in DSPy changes:
https://github.com/stanfordnlp/dspy/blob/dev_finetune/dspy/clients/anyscale.py

Fine tuning job working
Screenshot 2024-10-23 at 9 06 52 AM
Screenshot 2024-10-23 at 9 07 14 AM

Results on dev set and test set
Screenshot 2024-10-23 at 9 23 11 AM
Screenshot 2024-10-23 at 9 26 34 AM

Deployed Anyscale services:
Screenshot 2024-10-23 at 11 23 16 AM
Screenshot 2024-10-23 at 11 23 27 AM


templates/e2e-dspy-workflow/.gitignore
```diff
@@ -0,0 +1,3 @@
+*_checkpoints*/
```
isaacbmiller, line 1: File to be removed

templates/e2e-dspy-workflow/README.ipynb
```diff
@@ -0,0 +1,1074 @@
...
+    "     - Pipeline parallel size: 2\n",
+    "     - Tensor parallelism: 2\n",
+    "\n",
+    "3. Alternative setup (8x24GB GPUs):\n",
+    "   - Pipeline parallel size: 1\n",
+    "   - Tensor parallelism: ceil(150/24) = 7\n",
+    "\n",
+    "This configuration allows us to run the 70B model efficiently across our available GPU resources.\n",
+    "\n",
+    "Note that on Anyscale, you CANNOT download a 70B model without changing HF_HOME on most machines. The folder `/mnt/local_storage/' has enough space for a model download. It is not persisted across cluster restarts, but that is fine for model weights we don't need to save."
```
SumanthRH, line 510: Nit; I would rephrase this to say "On Anyscale Workspaces, we recommend to set ...."

templates/e2e-dspy-workflow/README.ipynb
```diff
@@ -0,0 +1,1074 @@
...
+    "\n",
+    "   ```\n",
+    "   vllm serve meta-llama/Meta-Llama-3.1-70B-Instruct --port 8000 --pipeline_parallel_size 2 --enable_prefix_caching --tensor_parallel_size 2\n",
+    "   ```\n",
+    "\n",
+    "## Parallelism Configuration\n",
+    "\n",
+    "We've chosen pipeline parallelism and tensor parallelism of 2 for the 70B model based on our current setup. Here's the reasoning:\n",
+    "\n",
+    "1. Model size: The 70B model has 30 parts of ~5 GB each (based on [HuggingFace documentation](https://huggingface.co/meta-llama/Meta-Llama-3.1-70B-Instruct/tree/main)).\n",
```
SumanthRH, line 493: Hmm this can even be removed since it's not a focus for the notebook. Or if oyu really want to keep it make it a dropdown widget (like our ray summit notebooks)

templates/e2e-dspy-workflow/README.ipynb
```diff
@@ -0,0 +1,1074 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Fine-tuning\n",
+    "\n",
+    "We will use LLM Forge to fine-tune the 1B model.\n",
+    "\n",
+    "In order to do this, we need to format our data into the correct format (Follows OpenAI messaging format).\n",
+    "\n",
+    "We can let DSPy do the rest, where it will properly generate the config and run the finetuning."
```
SumanthRH, line 643: This can be better highlighted that DSPy comes with an `anyscale` client out of the box for fine-tuning on Anyscale?

templates/e2e-dspy-workflow/README.ipynb
```diff
@@ -0,0 +1,1074 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Evaluation\n",
+    "\n",
+    "## Performance comparisons\n",
+    "\n",
```
SumanthRH, line 703: Probably better to add some context here for why you chose these baselines?

templates/e2e-dspy-workflow/README.ipynb
```diff
@@ -0,0 +1,1074 @@
...
+    "from dspy.clients.lm import TrainingMethod\n",
+    "from src import finetuning_kwargs\n",
+    "\n",
+    "train_data = dataset_formatted\n",
+    "method = TrainingMethod.SFT\n",
+    "\n",
+    "finetuneable_lm = dspy.LM(model=\"meta-llama/Llama-3.2-1B-Instruct\")\n",
+    "\n",
+    "try:\n",
+    "    finetuning_job = finetuneable_lm.finetune(method, train_data, eval_data=None, provider=\"anyscale\", train_kwargs=finetuning_kwargs)\n",
```
SumanthRH, line 689: As discussed offline: I'm a little concerned about the setup here.  Looks like the `anyscale` client is using the models SDK to get the `storage_uri` field and then get model checkpoints. Currently, with `llmforge>=0.5.6` this field is gonna actually contain the path to the forwarded checkpoint according to eval loss. This is not ideal, but I also want to make sure that internal DSPy code doesn't break with a simple upgrade. 

Can we just have a manual step for retrieving the checkpoint path? It's in fact also logged to the exp. tracking tool (wandb) so we can also have a util here that can retrieve it from the run.

SumanthRH (commented):
Left some comments. Can you also provide a link (for discussion only) to the DSPy branch this code is depending on for the short term, and ideally just the anyscale client related files? I'm interested in taking a look. 

I don't think there's much to comment on the content here since it's just regular data processing + fine-tuning, so most comments are nits.

templates/e2e-dspy-workflow/README.ipynb
SumanthRH: Like is it the standard in DSPy fine-tuning to have this `result` object that contains information for all the checkpoints?

SumanthRH: My expectation was that the `.finetune` client would just interface with DSPy datasets and then submit a fine-tuning job. But maybe I'm missing context on the common API

templates/e2e-dspy-workflow/configs/training/full_param/llama-3-8b.yaml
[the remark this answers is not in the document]
isaacbmiller: Design decision i guess. In case people want to directly modify this workspace instead of copying the changes over to the FT template

templates/e2e-dspy-workflow/src/utils.py
```diff
@@ -0,0 +1,115 @@
...
+
+class IntentClassificationModule(dspy.Module):
+    def __init__(self, labels_in_use):
+        self.intent_classifier = dspy.ChainOfThought(IntentClassification)
+        self.valid_labels = set(labels_in_use)
+
+    def forward(self, text):
+        prediction = self.intent_classifier(intent=text)
+        sanitized_prediction = dspy.Prediction(label=prediction.label.lower().strip().replace(" ", "_"), reasoning=prediction.reasoning)
+        return sanitized_prediction
```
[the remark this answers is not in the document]
isaacbmiller: Done this way because I dont know any easy ways to import modules from the main notebook. Could always pass around the class references in the notebook.

templates/e2e-dspy-workflow/src/utils.py
```diff
@@ -0,0 +1,115 @@
...
+        "DSP_CACHEDIR",
+        "HF_TOKEN",
+        "HF_HOME"
+    ]
+
+    for var in necessary_env_vars:
+        assert os.environ[var], f"{var} is not set"
+
+# -------------------------------
+MAX_TOKENS = 1000
```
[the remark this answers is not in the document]
isaacbmiller: ack

templates/e2e-dspy-workflow/README.ipynb
isaacbmiller: common API is wip. The way it currently works:
job.result() has the last (or best) checkpoint, but any time we fine-tune in DSPy we want to go over the entire list of checkpoints to find the best, so calling .result() in order to block and wait and then getting job.model_names is the way to get all the possible checkpoints

templates/e2e-dspy-workflow/README.ipynb
isaacbmiller: but it will crash if you don't change it

SumanthRH: Yeah but the wording is so negative here. Templates are user-facing and so you can expect to be one of the first points of contact with the platform. I would recommend better wording: 

"Note that by default, the cache directory used by HuggingFace is in the home directory (`/home/ray` here). We'll use `/mnt/local_storage` here for downloading large model weight files..."

templates/e2e-dspy-workflow/README.ipynb
SumanthRH, line 1: Okay so the file seems to be too big for comments in the UI, so I'm viewing this locally and writing up suggestions.


I'm leaving a first batch of comments : 

- DSPy installation: It looks like `dev_finetune` branch does not exist anymore? Is the auto-deletion for this merged branch a mistake?
https://github.com/stanfordnlp/dspy/pull/1594
a. Could we just install from `main` branch now and pin to the correct commit? 
b. For the installation, I'd recommend placing it in a `/tmp` or the `/home/ray` folder because there can be import issues when the dspy package is placed as  a subfolder in the current location (You can try it out yourself) 

- Dependencies: Can you please think about what the final image this template will use (guessing it's `aica-endpoints` ) and then think about what additional dependencies your code needs? I'm seeing `matplotlib` atleast that needs to be installed in the beginning. 
- Regarding environment variables: Could you provide a code block as follows (maybe also use the "replace" tag here) :
```
import os
os.environ["HF_HOME"] = "/mnt/local_storage/.cache/huggingface" 
# Make sure to uncomment if you haven't set the below env vars
# os.environ["HF_TOKEN"] = "enter token here"
# os.environ["WANDB_API_KEY"] = "enter token here"
```
- Have you done a full E2E run for the notebook in a template ? I'm seeing multiple minor errors when I do so.
- In "Gathering training data and running the 70B Model" is the command meant to be run in the terminal? I believe the `export HF_HOME=..` is missing in the markdown cell but present in the code cell.

image

  a. If the command is meant to be run in the terminal, maybe mention that env vars are best added in the "Dependencies" tab? Otherwise you'll get the hf token errror.
- Compute config for this template is weird? I'm not sure if we have a setup for templates to use gpu-only head node (in fact that throws up a couple warnings right away). Isn't it meant to be the same as the other templates - 8CPU head node with auto-select worker nodes? Why can't we just use `rayllm` to do `serve run rayllm_config_70B.yaml`  here? It'll be zero setup and as good as running things locally. We don't have to go through a rayllm tutorial here, we can revisit later.  Let me know if you need help with this. 
- Nit: Why is the `dspy` install editable lol

isaacbmiller: > DSPy installation: It looks like dev_finetune branch does not exist anymore? Is the auto-deletion for this merged branch a mistake?
> https://github.com/stanfordnlp/dspy/pull/1594
> a. Could we just install from main branch now and pin to the correct commit?
> b. For the installation, I'd recommend placing it in a /tmp or the /home/ray folder because there can be import issues when the dspy package is placed as a subfolder in the current location (You can try it out yourself)
Pinned to a release from pypi now

> Dependencies: Can you please think about what the final image this template will use (guessing it's aica-endpoints ) and then think about what additional dependencies your code needs? I'm seeing matplotlib atleast that needs to be installed in the beginning.
It is aica-endpoints.
The only non-dspy dependency is matplotlib, and I added that as an install in the beginning.

> Regarding environment variables: Could you provide a code block as follows (maybe also use the "replace" tag here) :
> 
> import os
> os.environ["HF_HOME"] = "/mnt/local_storage/.cache/huggingface" 
> # Make sure to uncomment if you haven't set the below env vars
> # os.environ["HF_TOKEN"] = "enter token here"
> # os.environ["WANDB_API_KEY"] = "enter token here"
Ack, done

> Have you done a full E2E run for the notebook in a template ? I'm seeing multiple minor errors when I do so.
Did a full clean run through using the RayLLM 70B changes.

> In "Gathering training data and running the 70B Model" is the command meant to be run in the terminal? I believe the export HF_HOME=.. is missing in the markdown cell but present in the code cell.
RayLLM fixes this

> a. If the command is meant to be run in the terminal, maybe mention that env vars are best added in the "Dependencies" tab? Otherwise you'll get the hf token errror.
Added due to the potential notebook bug we briefly discussed offline, I also recommend that people add the env vars directly to the notebook.

> Compute config for this template is weird? I'm not sure if we have a setup for templates to use gpu-only head node (in fact that throws up a couple warnings right away). Isn't it meant to be the same as the other templates - 8CPU head node with auto-select worker nodes? Why can't we just use rayllm to do serve run rayllm_config_70B.yaml here? It'll be zero setup and as good as running things locally. We don't have to go through a rayllm tutorial here, we can revisit later. Let me know if you need help with this.
Solved offline. Everything is now using RayLLM and on a CPU head node with auto-select worker nodes

> Nit: Why is the dspy install editable lol
Ack. Fixed

templates/e2e-dspy-workflow/.env.example
isaacbmiller, line 1: To be removed because this is solved inside of the notebnook

templates/e2e-dspy-workflow/README_files/README_56_0.png
isaacbmiller, line 1: to be removed

templates/e2e-dspy-workflow/configs/training/lora/llama-3-8b.yaml
```diff
@@ -1,6 +1,4 @@
-model_id: meta-llama/Meta-Llama-3-8B-Instruct # <-- change this to the model you want to fine-tune
-train_path: s3://llm-guide/data/viggo/train.jsonl # <-- change this to the path to your training data
-valid_path: s3://llm-guide/data/viggo/val.jsonl # <-- change this to the path to your validation data. This is optional
+model_id: meta-llama/Meta-Llama-3-8B-Instruct # <-- change this to the model you want to fine-tunetrain_path: s3://llm-guide/data/viggo/train.jsonl # <-- change this to the path to your training data
```
SumanthRH, line 1: is this intended? `train_path` should be in the next line?

isaacbmiller: Both removed because there technically should be no default (valid_path wasn't getting set to None, so eval loss on WANDB was from viggo

SumanthRH (approved):
LFG!

PR #385, opened 2024-10-23
Pin dependencies for DSPy template

isaacbmiller:


SumanthRH:
could you also clear out cell numbers btw using the same `clear_cell_nums.py` script from the E2E llm workflows template please?

SumanthRH (approved):

PR #386, opened 2024-10-24
Clarify service URLs and remove unused auth token

RCdeWit:
The auth token doesn't seem to be necessary to actually query the service. I've also added a line that makes explicit which API endpoint is being used.


Aydin-ab:

PR #391, opened 2024-11-08
Update templates after v0.5.8 `llmforge` release 

SumanthRH:
# What does this PR do?

Updates workspace templates after v0.5.8 release of `llmforge` . Product release has happened already so this PR can be safely merged. 

Some important changes in this version:
- `checkpoint_every_n_epochs` is deprecated in favour of `checkpoint_and_evaluation_frequency`
- `max_num_checkpoints` is deprecated. 
- `awsv2` -> `aws` in `RemoteStoragePath`. This is because of awsv2 cli deprecation. Since RemoteStoragePath is a bleeding edge feature, this is hard deprecation. 
- `TorchCompileConfig` is here. 

Also, we added liger support in the previous release - 0.5.7 but that was not added until now. This PR also adds liger to our configs. 

FWIW, torch compile + liger has some subtleties around compatibility - the best configuration for perf is turning on all liger flags, so this PR only adds liger to the configs. 

We've added liger only to the Lora configs since it's been hard to test out with A100s for full param.

### DIFF
--- a/templates/llm-router/README.ipynb
+++ b/templates/llm-router/README.ipynb
@@ -111,7 +111,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -306,7 +306,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -330,7 +330,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -461,7 +461,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 10,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -578,7 +578,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 11,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -644,7 +644,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 13,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -761,7 +761,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
@@ -904,7 +904,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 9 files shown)


SumanthRH:
Screenshots after testing Liger (from @erictang000 ) : 

image

templates/fine-tune-llm_v2/training_configs/custom/meta-llama/Meta-Llama-3-70B/lora/8xA100-80G-4k.yaml
SumanthRH, line 1: @kouroshHakha just wanna highlight that this is direct edit to the existing config. 

I think having a separate config with liger enabled is also doable, but given that we've tested out liger extensively regarding correctness, I'm fine with having this be in the defaults to squeeze out more performance - A lot of optionality is also confusing to the user.

kouroshHakha (commented):
Let's make liger versions a separate config. Mostly because there is an initialization time involved and it's not that faster either. So overall it's slower for 70B at least. 


image


I also didn't notice where compile params are. Where are they?

templates/fine-tune-llm_v2/training_configs/custom/meta-llama/Meta-Llama-3-8B/lora/4xA10-512.yaml
```diff
@@ -41,6 +41,16 @@ deepspeed:
...
+# Liger kernel configuration
+liger_kernel:
+  enabled: True
+  kwargs:
+    rms_norm: True
+    rope: True
+    swiglu: True
+    cross_entropy: True
+    fused_linear_cross_entropy: False
+
```
SumanthRH, line 53: @erictang000 umm did this value change? why was this false again?

templates/llm-router/README.ipynb
```diff
@@ -1400,6 +1402,25 @@
     "This plot illustrates that as we relax the cost constraints (i.e., increase the percentage of GPT-4 calls), the performance improves. While the performance of a random router improves linearly with cost, our router achieves significantly better results at each cost level."
    ]
   },
+  {
```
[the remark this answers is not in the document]
SumanthRH: I updated the router template to use the new 0.5.8 image. I noticed that the cell execution numbers are all messed up in the notebook, so I copied over some cleanup code from the E2E LLM Workflows template to cleanup cell nums and cached checkpoints.

templates/fine-tune-llm_v2/training_configs/custom/meta-llama/Meta-Llama-3-8B/lora/4xA10-512.yaml
erictang000: discussed on slack, for lower context length + batch size, regular cross entropy can be faster and memory is similar. Example w/ this toggled for llama-3.2-1b. But easier to just use defaults w/ liger, so we can turn fused linear cross entropy on in our default configs.
image
image

erictang000 (approved):
lgtm

kouroshHakha (approved):

PR #392, opened 2024-11-18
[RLlib] Introduce RLlib Intro template

ArturNiederfahrenhorst:
This PR introduces a template for RLlib.

Context -> https://docs.google.com/document/d/1UxNRSnOZ7QkZi1sw2e5mpSZ9xg-hBQw0rMzmuUe_MA0/edit?tab=t.0#heading=h.a084zdxe2jqv


templates/intro-rllib/atari_ppo.py
```diff
@@ -0,0 +1,100 @@
+import gymnasium as gym
+
+from ray import air
+from ray import tune
+from ray.rllib.algorithms.ppo import PPOConfig
+from ray.rllib.connectors.env_to_module.frame_stacking import FrameStackingEnvToModule
+from ray.rllib.connectors.learner.frame_stacking import FrameStackingLearner
+from ray.rllib.core.rl_module.default_model_config import DefaultModelConfig
+from ray.rllib.env.wrappers.atari_wrappers import wrap_atari_for_new_api_stack
+from ray.rllib.utils.test_utils import check_train_results_new_api_stack
```
sven1977, line 10: nit: not needed

templates/intro-rllib/atari_ppo.py
```diff
@@ -0,0 +1,100 @@
+import gymnasium as gym
```
sven1977, line 1: Is there any way, we can re-use the existing (tuned) example script?
Maybe similar to how the rst docs do it? Include certain blocks of the script, one at a time, into the `ipynb`? This would be nice, then we also won't have to worry about the template being tested, b/c it's already tested on the CI.

ArturNiederfahrenhorst: For any given release for which we want to update, we change the RAY_VERSION in 

```
!RAY_VERSION=2.39.0 # Set this when releasing an image instead of hardcoding it here
!curl -O https://raw.githubusercontent.com/ray-project/ray/refs/heads/releases/${RAY_VERSION}/rllib/tuned_examples/ppo/atari_ppo.py"
```

My idea here: That line goes into some file in a utility folder that we just live in the template.
No harm if users explore and find it. And when we want to update the template (which may happen often), we can use that small script every time to grab the newest version of RLlib's files that we need.

scottsun94:
cc: @aslonnie as well

scottsun94:
cc: @aslonnie @GokuMohandas  as well

templates/intro-rllib/atari_ppo.py
```diff
@@ -0,0 +1,100 @@
...
+        model_config=DefaultModelConfig(
+            conv_filters=[[16, 4, 2], [32, 4, 2], [64, 4, 2], [128, 4, 2]],
+            conv_activation="relu",
+            head_fcnet_hiddens=[256],
+            vf_share_layers=True,
+        ),
+    )
+)
+
+config.resources(num_gpus=0)
```
sven1977, line 88: this line is also not needed (old API stack setting)

templates/intro-rllib/atari_ppo.py
```diff
@@ -0,0 +1,100 @@
...
+config.learners(
```
sven1977, line 89: Let's move these into the main config bracket above.

templates/intro-rllib/atari_ppo.py
```diff
@@ -0,0 +1,100 @@
...
+    num_learners=4,
+    num_gpus_per_learner=1,
+)
+
+config.env_runners(num_env_runners=95)
```
sven1977, line 94: this, too (move these into the main config bracket above).

templates/intro-rllib/atari_ppo.py
```diff
@@ -0,0 +1,100 @@
...
+            "repeat_action_probability": 0.0,
+        },
+        clip_rewards=True,
+    )
+    .env_runners(
+        env_to_module_connector=_make_env_to_module_connector,
+    )
+    .training(
+        learner_connector=_make_learner_connector,
+        train_batch_size_per_learner=4000,  # 5000 on old yaml example
```
sven1977, line 66: remove: "# 5000 on old yaml example"

templates/intro-rllib/atari_ppo.py
```diff
@@ -0,0 +1,100 @@
...
+        minibatch_size=128,  # 500 on old yaml example
```
sven1977, line 67: remove: "# 5000 on old yaml example"

templates/intro-rllib/atari_ppo.py
ArturNiederfahrenhorst: Done

templates/intro-rllib/atari_ppo.py
ArturNiederfahrenhorst: Note: With the approach we are establish now as per our last offline conversation, this won't work anymore, since we copy the config from the tuned example from the ray repo.

Aydin-ab:

PR #395, opened 2025-01-03
Change hyperlink for llmforge to docs instead of versions table

SumanthRH:
# What does this PR do?

Tiny PR to fix a hyperlink. `LLMForge` should point to the intro page in the docs rather than the versions table.


erictang000 (approved):
💯

PR #401, opened 2025-03-15
[Data][LLM] Add template for ray.data.llm

lk-chen:
cc @kouroshHakha @comaniac

### DIFF
--- a/templates/ray-data-llm/util/utils.py
+++ b/templates/ray-data-llm/util/utils.py
@@ -1,0 +1,59 @@
+import os
+import requests
+
+import huggingface_hub
+
+HF_TOKEN_CACHE_PATH = "/mnt/local_storage/data/cache/huggingface/token"
+HF_TOKEN_LOCAL_PATH = "huggingface_token.txt"
+
+def is_on_gcp_cloud() -> bool:
+    """Detects if the cluster is running on GCP."""
+    try:
+        resp = requests.get("http://metadata.google.internal")
+        return resp.headers["Metadata-Flavor"] == "Google"
+    except:  # noqa: E722
+        return False
+
+
+def prompt_for_hugging_face_token(hf_model: str) -> str:
+    """Prompts the user for Hugging Face token if required by the model.
+    Returns the token as a string. If a token is not required by the model,
+    returns an empty string."""
+
+    url = f"https://huggingface.co/api/models/{hf_model}"
+
+    response = requests.get(url)
+    response_json = response.json()
+    is_private = response_json.get("private", False)
+    # possible gated response values: [False, "auto", "manual"]
+    is_gated = response_json.get("gated", False) in ("auto", "manual")
+    if response.status_code == 200 and not (is_private or is_gated):
+        return ""
+    elif response.status_code == 401 or (is_private or is_gated):
+        if os.path.isfile(HF_TOKEN_LOCAL_PATH):
+            return read_hugging_face_token_from_cache(HF_TOKEN_LOCAL_PATH)
+        if not os.path.isfile(HF_TOKEN_CACHE_PATH):
+            print("No cached Hugging Face token found. Starting authentication on VS Code overlay...")
+            # Starts authentication through VSCode overlay.
+            # Token saved to `HF_TOKEN_CACHE_PATH`
+            huggingface_hub.interpreter_login()
+
+        tkn = read_hugging_face_token_from_cache(HF_TOKEN_CACHE_PATH)
+        # Write the token to a local file, so it can be used by Ray job.
+        with open(HF_TOKEN_LOCAL_PATH, "w") as file:
+            file.write(tkn)
+        return tkn
+    else:
+        raise Exception(f"Failed to access the model. Status code: {response.status_code}")
+
+
+def read_hugging_face_token_from_cache(path) -> str:
+    try:
+        with open(path, "r") as file:
+            tkn = file.read()
+        print(f"Successfully read cached token at {path}.")
+        return tkn
+    except FileNotFoundError:
+        raise FileNotFoundError(
+            f"Could not find Hugging Face token cached at {path}."
+        )
--- a/templates/ray-data-llm/README.ipynb
+++ b/templates/ray-data-llm/README.ipynb
@@ -1,0 +1,238 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# LLM offline batch inference with Ray Data LLM APIs\n",
+    "\n",
+    "**⏱️ Time to complete**: 10 min\n",
+    "\n",
+    "\n",
+    "<!-- TODO: add a link for the API reference -->\n",
+    "This template shows you how to run batch inference for LLMs using Ray Data LLM.\n",
+    "\n",
+    "**Note:** This tutorial runs within a workspace. Review the `Introduction to Workspaces` template before this tutorial.\n",
+    "\n",
+    "\n",
+    "### How to decide between online vs offline inference for LLM\n",
+    "Online LLM inference (e.g. Anyscale Endpoint) should be used when you want to get real-time response for prompt or to interact with the LLM. Use online inference when you want to optimize latency of inference to be as quick as possible.\n",
+    "\n",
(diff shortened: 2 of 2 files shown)


kouroshHakha:
@kunling-anyscale Can you review this as well?

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+# For example, you can use ray.data.read_json(dataset_file) to load dataset in JSONL.
+
+df = datasets.load_dataset("cnn_dailymail", "3.0.0")
+ds = ray.data.from_huggingface(df["train"])
+
+```
+
+## Step 2: Define the processor config for vLLM engine
+
+We will also need to define the model configs for the LLM engine, which configures the model and compute resources needed for inference.
```
kouroshHakha, line 43: Can we just the tell the user to put their HF tokens in the cluster wide runtime envs?

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+```python
+# Prompts the user for Hugging Face token if required by the model.
+from util.utils import prompt_for_hugging_face_token
+HF_TOKEN = prompt_for_hugging_face_token("meta-llama/Meta-Llama-3.1-8B-Instruct")
+```
+
+In this example, we will be using the `meta-llama/Meta-Llama-3.1-8B-Instruct` model.
+We will also need to define a configuration associated with the model we want to use to configure the compute resources, engine arguments and other inference engine specific parameters. For more details on the the model configs, see the [API doc](https://docs.anyscale.com/llms/serving/guides/bring_any_model/) on bringing your own models.
```
kouroshHakha, line 58: This bring you own model thing will be removed, and we need to reference either an oss doc or address it here.

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Step 3: Define the preprocess and postprocess lambda
+
+
+We will need to define the preprocess lambda to convert input dataset to format that `vLLMEngineProcessor` can consume, and also postprocessor
+lambda that filter out the uninterested fields from vLLM engine.
+
+
+```python
+preprocess = lambda row: dict(
```
kouroshHakha, line 97: maybe define them as def instead of lambda?

templates/ray-data-llm/README.md
comaniac: +1. This aligns more on the real deployment case as well.

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+        max_model_len=16384,
+        enable_chunked_prefill=True,
+        max_num_batched_tokens=2048,
+    ),
+    # Override Ray's runtime env to include the Hugging Face token. Ray is being used under the hood to orchestrate the inference pipeline.
+    runtime_env=dict(
+        env_vars=dict(
+            HF_TOKEN=HF_TOKEN,
+        ),
+    ),
```
comaniac, line 82: Echo the previous comment, we could just say "you need to put your HF token here if the model you're using is gated".

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+    batch_size=16,
+    accelerator_type=accelerator_type,
```
comaniac, line 84: I'd suggest add `concurrency=1` and comment that you can scale out easily by incrementing this number.

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+)
+
+```
+
+## Step 3: Define the preprocess and postprocess lambda
```
comaniac, line 89:
```suggestion
## Step 3: Define the preprocess and postprocess functions
```

templates/ray-data-llm/README.md
comaniac: +1. If define separately, it's better and more flexible to be functions.

We should mention the input and output of preprocess and postprocess functions. Specifically, preprocess should be `preprocess(row: dict[str, Any])`, and its output should be a dict with keys `messages` and `sampling_params`. Postprocess should be `postprocess(row: dict[str, Any])`, and the `row` will have the key `generated_text`.

Also we should explain the column carry over mechanism. Specifically, the output rows of `preprocess` can only include the required keys (`messages` and `sampling_params`), but the output rows of `postprocess` should include all columns users want to carry over, including the columns in the input datasets.

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+from ray.data.llm import build_llm_processor
+
+processor = build_llm_processor(
+    processor_config,
+    preprocess=preprocess,
+    postprocess=postprocess,
+)
+
+ds = processor(ds)
+ds = ds.materialize()
```
comaniac, line 138: Add a comment saying that this could be any data writing APIs (e.g. `write_parquet`) and point to the API doc.

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+# Peak the first 3 entries. 
+sampled = ds.take(3)
+print("==================GENERATED OUTPUT===============")
+print('\n'.join(sampled))
+```
+
+
+### Monitoring the execution
+
+RayLLM-Batch uses Ray Data to implement the execution of the batch inference pipeline, and one can use the Ray Dashboard to monitor the execution. In the Ray Dashboard tab, navigate to the Job page and open the "Ray Data Overview" section. Click on the link for the running job, and open the "Ray Data Overview" section to view the details of the batch inference execution:
```
comaniac, line 150:
```suggestion
We can use the Ray Dashboard to monitor the execution. In the Ray Dashboard tab, navigate to the Job page and open the "Ray Data Overview" section. Click on the link for the running job, and open the "Ray Data Overview" section to view the details of the batch inference execution:
```

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+<img src="https://raw.githubusercontent.com/anyscale/templates/main/templates/batch-llm/assets/ray-data-jobs.png" width=900px />
+
+### Handling GPU out-of-memory failures
+If you run into CUDA out of memory, your batch size is likely too large. Set an explicit small batch size or use a smaller model (or a larger GPU).
```
comaniac, line 155: Better to point to the trouble shooting page (we should have that in the future).

templates/ray-data-llm/util/utils.py
```diff
@@ -0,0 +1,59 @@
...
+            # Token saved to `HF_TOKEN_CACHE_PATH`
+            huggingface_hub.interpreter_login()
+
+        tkn = read_hugging_face_token_from_cache(HF_TOKEN_CACHE_PATH)
+        # Write the token to a local file, so it can be used by Ray job.
+        with open(HF_TOKEN_LOCAL_PATH, "w") as file:
+            file.write(tkn)
+        return tkn
+    else:
+        raise Exception(f"Failed to access the model. Status code: {response.status_code}")
```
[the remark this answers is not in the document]
lk-chen: This function is removed as we ask user to always provide HF_TOKEN

templates/ray-data-llm/README.md
lk-chen: done, replaced with asking user to always provide HF_TOKEN

templates/ray-data-llm/README.md
```diff
@@ -0,0 +1,171 @@
...
+In this example, we will be using the `meta-llama/Meta-Llama-3.1-8B-Instruct` model.
+We will also need to define a configuration associated with the model we want to use to configure the compute resources, engine arguments and other inference engine specific parameters. For more details on the configs passed to vLLM engine, see [vLLM doc](https://docs.vllm.ai/en/latest/serving/engine_args.html).
+
+
+```python
+from ray.data.llm import vLLMEngineProcessorConfig
+from util.utils import is_on_gcp_cloud
+
+# There's no a10g on GCP.
+accelerator_type = "L4" if is_on_gcp_cloud() else "A10G"
```
kouroshHakha, line 61: L4 exist on both, right? Should we just remove this util all together by using L4 for both clouds?

kouroshHakha (commented):
just one last comment

comaniac (approved):

templates/ray-data-llm/README.md
lk-chen: Sounds good to me. I copied this from legacy template

lk-chen:
https://github.com/anyscale/templates/pull/404

PR #446, opened 2025-07-29
fixes for pretrain section of llama-factory template

erictang000:
update `c4_demo.json` -> `c4_demo.jsonl` and increase num epochs and decrease logging steps so wandb plot looks better.
image


SumanthRH (approved):
Thanks

PR #463, opened 2025-08-29
Getting Started with PyTorch Fully Sharded Data Parallel (FSDP2) and Ray Train

JasonLi1909:
A new workspace template that walks users through how to integrate PyTorch's FSDP2 with Ray Train. The purpose of this template is to allow customers to quickly get started with FSDP and Ray Train whether they are coming from using Distributed Data Parallel (DDP) or just getting started with training large models. 

For a high level overview, this template covers: 

- A hands-on example of training an image classification model
- Model checkpoint saving and loading with PyTorch Distributed Checkpoint (DCP)
- Configuring FSDP2 to mitigate out-of-memory (OOM) errors using mixed precision, CPU offloading, sharding granularity, and more
- GPU memory profiling with PyTorch Profiler
- Loading a distributed model for inference

**For easy testing:**
Simply copy the notebook into an Anyscale workspace (and preferably the image directory), provisioned with two T4 nodes.


templates/ray-train-fsdp/README.ipynb
marwan116, line 1: `PATH_TO_FULL_MODEL = "/mnt/cluster_storage/fsdp_mnist_16b0e0c2/full_model/full-model.pt"` hardcodes a hash that will not be the same when another user runs this notebook. Instead, either use a deterministic path or pathlib.Path to find the directory path.

marwan116, line 1: nitpick/preference: I usually prefer clear all cell outputs so the notebook is clean - if there are important outputs you want to preserve, then capture them in a markdown cell.

marwan116, line 1: on: CPU offloading and reshard_after_forward

You show a GPU memory profile titled "With CPU Offloading", what about reshard_after_forward ? I am assuming it is "With CPU Offloading and reshard_after_forward=True" correct ?

Also even if it is self evident from the chart, it is worth pointing out in text after the image that "the memory used by parameters drops significantly."

marwan116: Small preference but perhaps introduce reshard_after_forward below the "When not to use CPU offloading". You can also be perscriptive, when and when not to set reshard_after_forward=True...

You can state something like setting `reshard_after_forward=False` is roughly equivalent to `SHARD_GRAD_OP` in FSDP1 or ZeRO-2 in Deepspeed. State if the unsharded model parameters entirely fit on a single worker, then set it to False to avoid the additional communication.

marwan116, line 1: Perhaps worth pointing out under the mixed precision section is that because we are running with old GPUs T4s that don't support bfloat16 we choose a dtype of float16 - but generally this means you need to ensure your loss/gradients are scaled appropriately.

marwan116 (commented):
Nicely done guide on integrating with Ray Train with the latest state of PyTorch on FSDP2. I would also recommend we promote this material as part of our official Ray docs on Ray Train to ensure as much visibility and to share it with our open source community. I added some minor comments, hopefully you find them helpful.

JasonLi1909:
Closing as it was also made into an OSS Example

PR #600, opened 2026-04-10
Update rl skyrl template

kunling-anyscale:
The skyrl template is currently using an outdated Docker image and Ray version. We should update the template to align with the current installation guide: https://docs.skyrl.ai/docs/getting-started/installation.

I know we have used agent to migrate to the latest Ray version,but that isn’t the optimal approach. The skyrl template should be updated directly since it relies on a custom Docker image: novaskyai/skyrl-train-ray-2.51.1-py3.12-cu12.8.


Aydin-ab:
This was done in another PR 🤞

PR #605, opened 2026-04-16
updating paths in deployment serve llm template

elliot-barn:
updating paths in deployment serve llm template due to breaking ray doc build


Aydin-ab:
relative path won't work in the console

PR #608, opened 2026-04-17
[deps] Adding script to compile template dependencies - P0

elliot-barn:
**New top-level runner**                                                                       
  - update_deps.sh — downloads the raydepsets v0.0.1 binary (Linux x86_64 / Darwin arm64) and runs the compile 
  
**Updates**          
  - template.depsets.yaml — depset config pinning Ray 2.54.0 / Python 3.10, compiles ray_img_py310.lock                                                                         
  - scripts/fetch-ray-depsets.sh — fetches Ray's ray_img_py{version}.lock with fallback to requirements.txt                                                                           
  - scripts/fetch-ray-constraints.sh — fetches Ray's requirements_compiled_py{version}.txt   
  - depsets/ray_img_py310.lock — generated lockfile (2.3k lines)


dependencies/scripts/fetch-ray-constraints.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+RAY_VERSION="${1:?Usage: fetch-ray-constraints.sh <ray-version> <python-version>}"
+PYTHON_VERSION="${2:?Usage: fetch-ray-constraints.sh <ray-version> <python-version>}"
+
+DEST_DIR="/tmp/ray-deps"
+DEST_FILE="requirements_compiled_py${PYTHON_VERSION}.txt"
+BASE_URL="https://raw.githubusercontent.com/ray-project/ray/ray-${RAY_VERSION}/python"
```
aslonnie, line 9: this is fetching from the github source code? should we publish it somewhere instead, or do we have a longer term plan?

on the source code, we cannot really guarantee that it will be always saved at the same location.

elliot-barn: publishing them somewhere would be a great idea. We can publish to s3, doesn't feel right to publish as a github artifact anywhere. Any suggestions?

elliot-barn: given we are upgrading templates to keep them in sync or close to the latest release I think this is a fine implementation for now

templates/langchain-agent-ray-serve/python_depset.lock
```diff
@@ -0,0 +1,1526 @@
+--index-url https://pypi.org/simple
```
Aydin-ab: this looks like a forgotten artifact

Aydin-ab (approved):

PR #616, opened 2026-04-27
migrating tensor parallel autop

elliot-barn:
migrating tensor parallel autop from ray repo

Tested locally with rayapp test


templates/migrate-template.md
```diff
@@ -0,0 +1,273 @@
+---
```
Aydin-ab: you forgot to remove this file 🥸

Aydin-ab (approved):

PR #622, opened 2026-04-28
ci: /test-template supports multiple templates with parallel fan-out

Aydin-ab:
## What
Comment `/test-template foo bar baz` on a PR → fans out into one parallel Buildkite job per template (cap 3), each posting a clickable workspace URL annotation as soon as its workspace is created.

## Why
Today only one template can be tested per comment, and the URL it spawns isn't surfaced — you have to dig through Buildkite logs to find it.

## How

**GH Actions (`.github/workflows/test-template.yaml`)**
Parses `/test-template a b c` from the comment, validates names against `BUILD.yaml`, caps at 3, and dispatches one Buildkite build with `TEMPLATE_NAMES="a b c"`. Logic lives in `ci/parse-test-template-comment.sh`.

**Buildkite pipeline (`.buildkite/pipeline.template-test.yaml`)**
Thin bootstrap step: `ci/render-template-pipeline.sh | buildkite-agent pipeline upload`. The render script reads `TEMPLATE_NAMES` (space-separated) and emits one parallel step per template using Buildkite's dynamic pipeline pattern.

**Per-template watcher**
Each child step spawns a background watcher that grep's the always-printed line `Workspace created successfully id: expwrk_...`, fetches `cloud_id`/`project_id` via `anyscale workspace_v2 get -j`, and posts a Buildkite annotation:

> **`<template>`** workspace: `${ANYSCALE_HOST}/${CLOUD_ID}/${PROJECT_ID}/workspaces/${WS_ID}`

Annotation lands within seconds of workspace creation, well before the test finishes. Each annotation uses `--context "ws-<template>"`, so they coexist without overwriting.

## Manual Buildkite trigger
For one-off testing without the comment flow: New Build with these env vars:
```
TEMPLATE_NAMES=foo bar
ANYSCALE_CLI_TOKEN_SECRET_NAME=anyscale_cli_token_ci_staging
```

## Test plan
- [x] Local: parser handles 1/3-name (accept), 4-name (reject), `all` (rejected via BUILD.yaml lookup), shell-meta injection, empty
- [x] Manual build: 2 templates fanned out, all 2 annotations posted 
   - https://buildkite.com/anyscale/template-test/builds/130
- [ ] Post-merge: `/test-template a b` on a follow-up PR → both annotations visible on the build page

## Caveats
- GH Actions runs `issue_comment` workflows from the **default branch**, so the parser change only takes effect after merge. The Buildkite-side change is fully validated by manual builds against this branch.
- For "newer comment cancels older in-flight build", enable **Cancel Intermediate Builds** on the `template-test` pipeline (Buildkite UI → Pipeline Settings → Builds). UI-only.

### DIFF
--- a/.github/workflows/test-template.yaml
+++ b/.github/workflows/test-template.yaml
@@ -53,7 +53,7 @@ jobs:
           -d "$(jq -n \
             --arg branch "${{ steps.pr.outputs.branch }}" \
             --arg sha "${{ steps.pr.outputs.sha }}" \
-            --arg templates_json "${{ steps.parse.outputs.templates_json }}" \
+            --arg templates "${{ steps.parse.outputs.templates }}" \
             --arg msg "Test templates: ${{ steps.parse.outputs.templates }} (PR #${{ github.event.issue.number }})" \
             --arg secret "${{ secrets.ANYSCALE_CLI_TOKEN_SECRET_NAME }}" \
             '{
@@ -61,7 +61,7 @@ jobs:
               commit: $sha,
               branch: $branch,
               env: {
-                TEMPLATE_NAMES_JSON: $templates_json,
+                TEMPLATE_NAMES: $templates,
                 ANYSCALE_CLI_TOKEN_SECRET_NAME: $secret
               }
             }'
--- a/.github/workflows/test-template.yaml
+++ b/.github/workflows/test-template.yaml
@@ -12,18 +12,6 @@ jobs:
       contains(fromJSON('["MEMBER","OWNER","COLLABORATOR"]'), github.event.comment.author_association)
     runs-on: ubuntu-latest
     steps:
-    - name: Parse template name from comment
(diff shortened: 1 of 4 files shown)


.github/workflows/test-template.yaml
```diff
@@ -41,6 +29,42 @@ jobs:
           core.setOutput('branch', pr.data.head.ref);
           core.setOutput('sha', pr.data.head.sha);
 
+    - name: Checkout PR
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ steps.pr.outputs.sha }}
+
+    - name: Parse and validate template names
```
elliot-barn, line 37: can you add this to a script please (line 42-66)

Aydin-ab: Done in c5111382 ✅

elliot-barn (approved):

PR #623, opened 2026-04-28
ci: add BUILD.yaml validator (paths + GCP image reachability)

Aydin-ab:
## Summary
Validate `BUILD.yaml` and the compute config files it references. Split between **pre-commit** (cheap, offline) and **premerge** (network).

### Schema (pydantic, `extra="forbid"` at every level)

Every BUILD.yaml entry must have `name`, `dir`, `cluster_env`, `compute_config`, and `test`. Unknown keys are rejected anywhere they appear.

| Field | Rule |
|---|---|
| `name` | matches `^[a-z0-9_-]+$`, unique across the file |
| `dir` | starts with `templates/`, basename == `name` |
| `cluster_env` | exactly one of `image_uri` or `byod` (separate errors for has-both vs has-neither) |
| `cluster_env.image_uri` | starts with `anyscale/` |
| `cluster_env.byod` | requires `docker_image` + `ray_version` |
| `compute_config.GCP` | `^configs/.+/gce\.yaml$`, dir basename == `name` (or `basic-single-node`) |
| `compute_config.AWS` | `^configs/.+/aws\.yaml$`, same dir as GCP |
| `test.tests_path` | starts with `tests/`, basename == `name`, dir contains `tests.sh` |
| `test.command` | non-empty |
| `test.timeout_in_sec` | int > 0 |

The compute config files themselves are validated against the **legacy** [ComputeTemplateConfig][1] / [ComputeNodeType][2] / [WorkerNodeType][3] / [Resources][4] schemas (lax: any field optional, strict: no unknown keys). Catches files using the new schema.

[1]: https://docs.anyscale.com/ref/0.26.64/compute-config-api#computetemplateconfig-legacy
[2]: https://docs.anyscale.com/ref/0.26.64/compute-config-api#computenodetype-legacy
[3]: https://docs.anyscale.com/ref/0.26.64/other#workernodetype-legacy
[4]: https://docs.anyscale.com/ref/0.26.64/other#resources-legacy

### Filesystem checks
`dir`, both compute configs, `tests_path`, and `tests_path/tests.sh` must exist.

### Byod GCP image checks (`cluster_env.byod.docker_image`)
- **Canonical naming** (always): if hosted on the Anyscale registry (`us-docker.pkg.dev/anyscale-workspace-templates/workspace-templates/*`), the URI must equal `<registry>/<name>:<byod.ray_version>`. Catches drift between BUILD.yaml fields and what was published.
- **Reachability** (premerge only): for any GCP-hosted URI (`*.pkg.dev`, `gcr.io`), HEAD the registry manifest endpoint, require 200. The registry is public, so a 4xx almost always means the push didn't complete.

(`image_uri` is skipped — the schema constrains it to `^anyscale/`, never a GCP host.)

### Warnings (don't fail)
A custom compute config that's byte-equal to `configs/basic-single-node/{aws,gce}.yaml` warns the author to reuse the shared one.

## Where each check runs

| | Pre-commit hook | Premerge job |
|---|---|---|
| Schema + paths | ✅ | ✅ |
| Byod canonical naming | ✅ | ✅ |
| Byod registry reachability | ❌ | ✅ |

The pre-commit hook runs `validate_build_yaml.py --no-network` on `BUILD.yaml` changes, so authors get fast offline feedback.

## One-time content changes to land green

- **Config dir renames** (14): hyphens → underscores so each config dir's basename equals its entry's `name` (e.g., `biotech-boltz-screening` → `biotech_boltz_screening`).
- **`configs/llm-batch-inference/` split** into `configs/llm_batch_inference_text/` and `configs/llm_batch_inference_vision/` — was shared by two entries.
- **`configs/image-search-and-classification/`**: moved from inside its template dir up to `configs/` to match the convention.
- **`ray-data-batch-image-classification`**: switched to shared `basic-single-node` (its custom configs were byte-equal duplicates).
- **`ray-summit-core-masterclass`**: added a real `gce.yaml` — was reusing `basic-serverless-config/gce.yaml` while customizing AWS, which the new "GCP/AWS in same dir" rule disallows.
- **`basic-serverless-config` → `basic-single-node`**: 4 templates swapped (AWS files were byte-equal; GCP gets a small `n1-standard-8` → `n2-standard-8` instance upgrade).
- **`gcp.yaml` → `gce.yaml`**: renamed the one config that diverged from the GCP=`gce.yaml` convention.
- **TagSpecifications cleanup**: removed deprecated/dead feature tags `aws.TagSpecifications` blocks.
- **Republished 2 byod images under canonical names** (digests preserved via re-tag, no rebuild): `template_unstructured_data_ingestion:2.54.1` → `unstructured_data_ingestion:2.54.1`; `object-detection-ray:2.54.1` → `object-detection-video-processing:2.54.1`.

## Test plan
- [x] All 48 entries pass schema + paths + GCP image checks (naming + reachability)
- [x] All 71 referenced compute config files pass legacy schema validation at every nesting level
- [x] Local failure modes verified: missing path / unknown key at any level / private GCP image / new-schema config / canonical-naming mismatch / cluster_env has both/neither


Aydin-ab:
i'm enforcing `tests:` here but the "passthrough test" path is still available by giving an empty tests.sh
I wanna reduce the risk of an agent thinking tests are optional and lazily not making them

.github/workflows/premerge.yaml
```diff
@@ -18,12 +18,9 @@ jobs:
       with:
         python-version: '3.9'
 
-    # Install pre-commit dependencies (nbconvert provides the jupyter CLI
-    # that the existing generate-readme local hook needs)
     - name: Install pre-commit
-      run: pip install pre-commit==3.8.0 nbconvert==7.17.1
+      run: pip install pre-commit==3.8.0 nbconvert==7.17.1 pyyaml pydantic
```
elliot-barn: nit can you pin these please

elliot-barn (approved):
just the nit about pinning pyyaml and pydantic

.github/workflows/premerge.yaml
Aydin-ab: done

PR #627, opened 2026-04-30
Updating templates that are violating ray doc build rules

elliot-barn:
The Ray docs build (`-W --keep-going`) was failing on warnings emitted from template content fetched via sphinx-collections from the templates channel. Because those files are pulled from this repo at channel-build time, the fixes have to land in the template sources here. See `fixes.md` for the upstream warning details.

### 1. `templates/multi_agent_a2a/README.md`

Sixteen ` ```python ` code fences contained Jupyter shell-magic lines (`!python`, `!curl`, `!pip`, `!anyscale`). The Pygments python lexer cannot parse the leading `!`, producing `Lexing literal_block ... resulted in an error at token: '!'` warnings.

**Fix:** changed the fence language tag from `python` to `bash` for all sixteen blocks (lines 176, 216, 223, 230, 237, 244, 253, 263, 297, 304, 311, 318, 325, 343, 656, 684).

The six real-Python fences in this file (lines 284, 338, 529, 570, 630, 669) were left as `python`.

> Only lines 253 and 656 were flagged in the failing build, but Pygments falls back to "relaxed mode" after the first warning, masking the rest. Fixed all sixteen for consistency so the next channel build cannot regress here.

### 2. `templates/asynchronous_inference/README.md`

Two ` ```python ` fences contained shell magic — the Redis install block (line 57: `!sudo apt-get …`, `!redis-server …`, `!redis-cli …`) and the dependency install block (line 81: `!pip install …`).

**Fix:** changed both fence tags from `python` to `bash`. Only line 57 was flagged in the failing build; line 81 was fixed proactively for the same "relaxed mode" reason.

### 3. `templates/tune_pytorch_asha/README.md`

A `python` fence at line 108 mixed real Python (env-var setup, imports) with a trailing `!pip install -r requirements.txt`.

**Fix:** split the block in two — a `bash` fence with `!pip install -r requirements.txt` immediately above the Python `python` fence, matching the convention used by other templates in this repo. Also dropped the now-unused `sys, subprocess` imports from the Python block (they had only been there to support the install line).

### 4. `templates/llm_finetuning/notebooks/kto_lora.ipynb`

A markdown link in cell-4 was written as `[QLoRA]((https://...))` (double opening parens, double closing parens). MyST parsed the inner `(URL)` as a cross-reference target rather than the URL of a hyperlink, producing `'myst' cross-reference target not found: '(https://...)'`.

**Fix:** collapsed `]((` → `](` and `))` → `)` so MyST parses it as a plain external hyperlink. The URL itself was left unchanged (updating the `ray-project/ray` GitHub path is out of scope for the build break).


Aydin-ab:
/test-template asynchronous_inference llm_finetuning tune_pytorch_asha multi_agent_a2a

(need to pull master to make it work actually 😅)

elliot-barn:
/test-template asynchronous_inference llm_finetuning tune_pytorch_asha multi_agent_a2a

elliot-barn:
/test-template asynchronous_inference llm_finetuning tune_pytorch_asha

Aydin-ab:
the future pre-commit check enforcing using nbconvert to do .ipynb -> .md might just break this again (not sure) 😅 

if that's the case one solution is to make the change on ray docs side:
- make sphinx ignore these specific warnings on those files
- or modify them mid-build via hooks, you can see similar work i did here to work around sphinx `-W` compliance
    - https://github.com/ray-project/ray/pull/59984 
    - https://github.com/ray-project/ray/pull/59982
    - https://github.com/ray-project/ray/pull/59980

other solution is to relax that future pre-commit check but i'd rather favor strict README.ipynb == README.md rather than an invisible compliance rule on sphinx side

PR #690, opened 2026-05-13
ci: trigger tmpl-publish Buildkite pipeline on 'publish' label

Aydin-ab:
Applying the `publish` label triggers the `tmpl-publish` Buildkite pipeline and posts a commit status on the PR's HEAD sha linking to the build. This workflow posts the status to the templates PR explicitly (like `tmpl-test`).

After merge: `gh label create publish --description 'Trigger tmpl-publish Buildkite pipeline' --color 0E8A16`

### DIFF
--- a/.github/workflows/publish-on-label.yaml
+++ b/.github/workflows/publish-on-label.yaml
@@ -49,10 +49,12 @@ jobs:
       env:
         BUILD_URL: ${{ steps.buildkite.outputs.web_url }}
         PR_SHA: ${{ github.event.pull_request.head.sha }}
+        REPO: ${{ github.repository }}
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
+        # state=success reflects "trigger succeeded" — publish completion is gated manually downstream.
         gh api \
-          "repos/${{ github.repository }}/statuses/${PR_SHA}" \
+          "repos/${REPO}/statuses/${PR_SHA}" \
           -f state=success \
           -f target_url="$BUILD_URL" \
           -f description="Click Details to drive publish gates in Buildkite" \
--- a/.github/workflows/publish-on-label.yaml
+++ b/.github/workflows/publish-on-label.yaml
@@ -1,0 +1,70 @@
+name: publish-on-label
+
+on:
+  pull_request:
+    types: [labeled]
+
+permissions:
+  pull-requests: write
+  contents: read
+
+jobs:
+  trigger-publish:
+    if: github.event.label.name == 'publish'
+    runs-on: ubuntu-latest
+    steps:
+    - name: Trigger Buildkite tmpl-publish build
+      id: buildkite
+      env:
+        BUILDKITE_API_TOKEN: ${{ secrets.BUILDKITE_API_TOKEN }}
+        BUILDKITE_ORG: ${{ secrets.BUILDKITE_ORG }}
+        PR_NUMBER: ${{ github.event.pull_request.number }}
+        PR_TITLE: ${{ github.event.pull_request.title }}
+      run: |
+        response=$(curl -sfS \
+          -X POST \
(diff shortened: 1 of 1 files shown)


Aydin-ab (commented):
**LGTM** — workflow is functionally correct and properly mitigates the shell-injection vector. Findings below.

## Checklist

1. **Trigger filter** ✅ — `on: pull_request: types: [labeled]` + job-level `if: github.event.label.name == 'publish'`. Fires only when the just-added label is `publish`.
2. **Permissions block** ✅ — `pull-requests: write` + `contents: read`, nothing else.
3. **Buildkite API call shape** ✅ — matches `test-template.yaml`: bearer auth, `https://api.buildkite.com/v2/organizations/${BUILDKITE_ORG}/pipelines/tmpl-publish/builds`, jq for body, `curl -sfS` so HTTP failures exit non-zero. The added `response`/`web_url`/null-check around the curl output is a good extension over the prior art (test-template discards the body via `-o /dev/null`).
4. **Secrets** ✅ — reuses `BUILDKITE_API_TOKEN` and `BUILDKITE_ORG`; pipeline slug `tmpl-publish` is hardcoded (no new secret).
5. **Shell injection** ✅ — `PR_TITLE` and `PR_NUMBER` flow through `env:` and are referenced as `${PR_TITLE}`/`${PR_NUMBER}` inside a double-quoted string passed to `jq --arg`. A malicious title containing shell metacharacters (semicolons, backticks, `$(...)`) is inserted verbatim into jq's string value — bash does not re-tokenize expanded variables, and `jq --arg` treats its value as a literal. `BUILD_URL` and `PR_BRANCH` are likewise read via `process.env.*` in the github-script step, not via `${{ }}` interpolation in the script body.
6. **PR comment hint** ✅ — body explicitly includes `tmpl-branch=${prBranch}` sourced from `github.event.pull_request.head.ref`.
7. **No self-trigger loop** ✅ — workflow only posts a comment; commenting doesn't fire `labeled`.
8. **Label-removal handling** ✅ — `types: [labeled]` only (not `[labeled, unlabeled]`).
9. **Fork threat model** ✅ — only users with triage+ can apply labels; `pull_request` (not `pull_request_target`) means fork-PR runs don't see secrets, so the curl would 401 before any harm.

## Nits (non-blocking)

- **NIT**: `contents: read` is unused — the workflow never checks out the repo. Could be dropped for stricter least-privilege, but it's a sane default and matches `test-template.yaml`, so feel free to leave it.
- **NIT**: No `concurrency:` group. If a maintainer removes-and-re-adds the `publish` label in quick succession, two builds fire in parallel. Adding `concurrency: { group: publish-pr-${{ github.event.pull_request.number }}, cancel-in-progress: false }` would coalesce duplicates. Low impact since the human is in the loop.
- **NIT**: `curl -sfS` exits non-zero on a Buildkite 4xx/5xx but drops the response body from the log, leaving a useless `curl: (22) ...` error. Consider `curl --fail-with-body -sS ...` (curl ≥7.76, available on `ubuntu-latest`) so any Buildkite error message lands in the workflow log.

Ship it.

Aydin-ab (commented):
All 3 NITs addressed in d9ceb1a1. Thanks for the review!

Aydin-ab (commented):
Updated in 687b13fd: switched from a PR comment to a GH commit status check so the result shows up in the PR check list (like `tmpl-test`). The pipeline lives in `anyscale/product` so Buildkite's native commit-status post can't reach the templates PR — this gets us there manually.

Aydin-ab:
Superseded by #706 (pre-merge drift-gated publishing). #706 uses a single required `buildkite/publish-templates` check that fires on every PR, computes drift per-template, and fans out to per-template publishes through prod when a `go` label is applied. Same intent as this PR (automate publishing) but the new design enforces strong consistency between `main` and prod and consolidates to one check.

PR #804, opened 2026-06-30
Fix unstructured_data_ingestion: pin accelerate for transformers 4.57…

elliot-barn:
….3 compat

The template pinned transformers==4.57.3 but let accelerate float to an incompatible version, causing a circular import in the runtime_env pip venv when `from transformers import pipeline` runs:

  ImportError: cannot import name 'dispatch_model' from partially
  initialized module 'accelerate.big_modeling' (circular import)

Pin accelerate==1.10.1 (verified to import transformers.pipeline cleanly with transformers==4.57.3) across all three install paths:
- runtime_env packages list (the path the probe traceback runs in)
- the README pip install cell
- the Dockerfile final RUN pip install

README.md regenerated via nbconvert. BYOD template (no python_depset.lock), so no raydepsets recompile.


Aydin-ab:
is it still relevant after the 2.56 bump ?

PR #835, opened 2026-07-01
ci(depsets): weekly scheduled lock refresh that opens a PR on drift

Aydin-ab:
A weekly job that recompiles all depset locks and opens a PR when they drift — so staying current with upstream is a deliberate, reviewable step, not a random tax on unrelated PRs.

## Why
`check-depsets` re-resolves locks against live indexes, so committed locks slowly drift from "what a fresh compile produces." #834 stops that from failing unrelated PRs; this is the other half — refresh the locks *on a cadence*, centrally, in one PR.

## What
`.github/workflows/refresh-depsets.yaml` — Mondays 08:00 UTC (+ manual dispatch):
- Runs the full `./update_deps.sh` (retried for transient index 503s).
- If any lock drifted, commits to a rolling `deps/scheduled-lock-refresh` branch and opens (or updates) a single refresh PR.

## Notes / caveats
- **Requires** the repo/org setting **"Allow GitHub Actions to create and approve pull requests"** for the `GITHUB_TOKEN` PR-open step; otherwise swap in a PAT.
- The refresh PR's own `check-depsets` should pass (it *is* a fresh compile). DCO: the bot commit is `-s` signed off as `github-actions[bot]` — confirm your DCO check accepts that.
- Independent of #831/#834; complements them.
- Not exercisable in this PR (workflows run only from `main`) — verify with a manual `workflow_dispatch` after merge.

### DIFF
--- a/.github/workflows/refresh-depsets.yaml
+++ b/.github/workflows/refresh-depsets.yaml
@@ -1,0 +1,63 @@
+name: refresh-depsets
+
+# Weekly full recompile of every depset lock against current package indexes.
+# This is where "keep up with upstream" happens — deliberately, in one reviewable
+# PR — instead of ambient drift failing unrelated PRs' check-depsets gate.
+# See .claude/skills/template/references/dependencies.md.
+#
+# NOTE: opening a PR from GITHUB_TOKEN requires the repo/org setting
+# "Allow GitHub Actions to create and approve pull requests" to be enabled.
+
+on:
+  schedule:
+    - cron: "0 8 * * 1"   # Mondays 08:00 UTC
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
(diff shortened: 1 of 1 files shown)


Aydin-ab:
Closing as not-now — superseded in spirit by #834's rework.

#834 now does **per-template scoping** (regenerate only a changed template's lock via `raydepsets --name` + diff), so the per-PR gate no longer re-resolves every lock and doesn't depend on a scheduled refresh to stay green. Actively-maintained templates already get their locks recompiled through the ~monthly Ray-bump fanout, and observed ambient drift is low (~2 lines/month). The residual value here — proactive CVE/freshness for dormant templates + base locks, and surfacing upstream breakage on a cadence — is real but marginal, and doesn't justify the external setup (repo "Allow Actions to create PRs" + DCO) right now.

Branch kept; reopen if dormant-lock drift or an upstream-breakage ambush actually bites.

PR #913, opened 2026-07-18
fix(configs): pin head unschedulable — train/tune (batch 6)

Aydin-ab:
Pins the head node unschedulable (`resources: {CPU: 0}`) in the aws + gce compute configs. These configs have worker groups or auto_select, so the head is a coordinator, not a work node: the launch-time conversion injects this when the field is absent, so setting it explicitly makes the repo config equal what ships to S3 (and `rayapp test` reproduces the real worker placement instead of co-locating on a schedulable head). Heads are CPU-only instances, so `CPU: 0` alone is unschedulable. Config-only.

Templates: tensor_parallel_dtensor, ray-tune-train-integration, tune_pytorch_asha, pytorch-profiling

## Testing
Validated green via `/test-template` (Buildkite). The earlier `CPU: 0, GPU: 0` → `CPU: 0` normalization is a no-op on CPU-only heads (identical Ray resources), so not re-tested.

Part of the head-unschedulable sweep; a companion check (#924) enforces this policy repo-wide.

https://claude.ai/code/session_01QmLc4yWtmC3PX2NwWzPbzD

### DIFF
--- a/configs/pytorch-profiling/aws.yaml
+++ b/configs/pytorch-profiling/aws.yaml
@@ -1,5 +1,7 @@
 head_node:
   instance_type: m5.2xlarge
+  resources:
+    CPU: 0
 worker_nodes:
 - name: 1xT4:4CPU-16GB
   instance_type: g4dn.xlarge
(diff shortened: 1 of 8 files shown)


Aydin-ab:
/test-template tensor_parallel_dtensor ray-tune-train-integration tune_pytorch_asha

elliot-barn (approved):

PR #916, opened 2026-07-19
fix(templates): deliver added deps via runtime_env (audio, e2e-timeseries, image-search)

Aydin-ab:
Wire added deps to workers/replicas via runtime_env so the standalone Service/Job path matches the workspace (per the /template skill Runtime skew guidance, #905). Part of the dependency good-practice pass.

## What changed
- audio-dataset-curation-llm-judge: add ray.init(runtime_env={pip: lock}) — it had none, so the resample UDF + Whisper/Llama actors got deps only via workspace auto-propagation.
- e2e-timeseries-forecasting (3 notebooks): add pip:lock to the existing runtime_env (was py_modules only), anchored to the package file.
- image-search-and-classification: version currency only (Ray 2.51→2.56, mlflow 2.19→2.22 across 3 configs + containerfile); ship-path deps already correctly scoped.
- numpy pin deferred to the in-flight migrate-to-multimodal branch.

## Testing
/test-template on the three templates. Note: /test-template runs the workspace path (auto-propagation covers it), so it won't fully exercise the standalone Service/Job path these fixes improve.

### DIFF
--- a/templates/audio-dataset-curation-llm-judge/README.ipynb
+++ b/templates/audio-dataset-curation-llm-judge/README.ipynb
@@ -7,11 +7,11 @@
     "# Audio batch inference\n",
     "\n",
     "<div align=\"left\">\n",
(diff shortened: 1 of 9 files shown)


Aydin-ab:
/test-template audio-dataset-curation-llm-judge e2e-timeseries-forecasting image-search-and-classification

elliot-barn (approved):

PR #932, opened 2026-07-30
fix(templates): deliver locked deps to Serve replicas (langchain, multi_agent_a2a, serve-sd-v2)

Aydin-ab:
Three Serve templates installed deps head-only (`uv pip --system`), which never reaches replicas that land on workers under the published head-unschedulable compute config — so a standalone Service 500s with ModuleNotFoundError. Each deployment now carries `runtime_env={"pip": python_depset.lock}` so replicas install the locked closure wherever they run.

## What changed
- **langchain-agent-ray-serve** — `runtime_env` on the WeatherMCP + LangGraphServeDeployment decorators.
- **multi_agent_a2a** — `pip: python_depset.lock` on all 8 non-LLM apps (llm app unchanged); dropped the redundant top-level `requirements:`; `tests.sh` now deploys in two passes and asserts a real agent call (an agent missing its deps would `DEPLOY_FAILED`).
- **serve-stable-diffusion-v2** — `runtime_env` on both deployments + a new lock-pinned `service_config.yaml`.

Clears the two probe-red Serve templates; `mcp-ray-serve`'s sibling fix is #931.

## Testing
Verified by diff review; `/test-template` not yet run. CI timeouts for multi_agent/serve-sd raised to 2700s — replicas now install the closure at startup (the tradeoff of `runtime_env` delivery).

https://claude.ai/code/session_01WQASugoLWZWj2Lp5AYRDka


Aydin-ab:
/test-template langchain-agent-ray-serve multi_agent_a2a serve-stable-diffusion-v2

Aydin-ab:
/test-template langchain-agent-ray-serve multi_agent_a2a serve-stable-diffusion-v2

elliot-barn (approved):

PR #939, opened 2026-08-07
fix(entity-recognition-with-llms): minimal deps + explicit worker delivery

Aydin-ab:
## What

The template hand-listed `llamafactory` + `transformers/peft/trl/omegaconf/torchdata` in three places and relied on workspace auto-propagation to reach the GPU training worker.

## Fix

LLaMA-Factory **main** pulls peft/trl/datasets/torchdata itself, so:
- **requirements.txt** is just LLaMA-Factory (main — no *release* supports recent transformers; 0.9.5 caps at 5.6) + `transformers<5` held on the **validated 4.x line** + the two extras it doesn't pull (`pynvml`, `tensorboard`).
- **run_training.py** delivers that closure to the `@ray.remote` GPU task via `runtime_env` instead of head-side propagation.
- **containerfile** mirrors the minimal set; fixed the README's stale containerfile link + `py311-cu124` image reference.

## Notes

- git-main LLaMA-Factory can't be hash-locked (git dep), so this template is **system-deps** by nature — a BYOD image is the ideal future delivery.
- Kept `transformers<5` (the tested line) rather than chasing the base image's newer build — the exact image transformers version isn't reliably knowable from the repo (its base *depset lock* drifts from the deployed image), so no speculative upgrade.

https://claude.ai/code/session_01WQASugoLWZWj2Lp5AYRDka

### DIFF
--- a/templates/entity-recognition-with-llms/README.ipynb
+++ b/templates/entity-recognition-with-llms/README.ipynb
@@ -69,7 +69,7 @@
     "<img src=\"https://raw.githubusercontent.com/anyscale/foundational-ray-app/refs/heads/main/images/compute.png\" width=500>\n",
     "\n",
     "### Dependencies\n",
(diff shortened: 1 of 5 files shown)


svc-template-updater (approved):
Build 607 green on `/test-template entity-recognition-with-llms`.

Approving for one reason above the others: the missing `runtime_env` on the `@ray.remote(num_gpus=1)` task is a real break. `configs/entity-recognition-with-llms/aws.yaml` sets head `CPU: 0`, so that task always lands on a GPU worker, and `llamafactory` is not in the stock `ray-llm` image the template actually runs (the containerfile is not wired into BUILD.yaml). As a Job or Service it fails on `llamafactory-cli` not found.

Two notes for later, neither blocking: the `LlamaFactory` → `LLaMA-Factory` URL change is cosmetic (both resolve to the same SHA via GitHub's redirect), and the new `requirements.txt` still wants a compiled lock + depsets entry seeded from `ray-llm-2.56.0-py312-cu130.freeze.txt`, or it is a second hand-maintained pin list.

PR #940, opened 2026-08-08
feat(templates): declarative K8S compute configs for 3 pilot templates

Aydin-ab:
Adds an optional `K8S:` compute-config key pointing at a declarative `k8s.yaml` — `required_resources` shapes resolved server-side into free pods, no registered instance types, one config covering AKS/EKS/GKE. Three pilots span the shape space: `workspace-intro` (single node), `getting-started` (multi-node CPU), `pytorch-fsdp` (multi-node GPU, 2×T4 via `required_labels: {ray.io/accelerator-type: T4}` — Azure GPU pools have no L4/A10G). The validator accepts the new key and enforces the declarative contract: no `instance_type`, no `auto_select_worker_config` (incompatible with declarative configs), and GPU counts paired with the accelerator-type label (launch validation reads the label, not a `required_resources.accelerator` key).

Groundwork for the Azure/K8s template path: `rayapp build` already carries the key into bundles as-is; the product launch path selecting + honoring it lands separately (today it force-selects AWS and flattens on K8s clouds), so the key is inert in prod until then.

## Testing
- `pre-commit run --all-files` green (BUILD.yaml validator gate included).
- `rayapp build`: all 56 bundles build; the three pilots' `ray-app.json` carry the `K8S` config with `required_resources` + `required_labels` preserved verbatim.
- All three files parse through `anyscale.compute_config.models.ComputeConfig` (`instance_type` → `custom`).
- Shapes checked against https://docs.anyscale.com/configuration/compute/declarative and the backend's `check_gpu_accelerator_consistency` (GPU count must pair with the `ray.io/accelerator-type` label; auto-select unsupported with declarative).
- Validator negative cases exercised: `instance_type`, `auto_select_worker_config`, GPU without label, `accelerator` inside `required_resources`, empty `required_resources`, non-dict nodes, cross-dir K8S path, wrong filename — each fails with a targeted error.

https://claude.ai/code/session_013BsS5dSk6s6xsLx2FbBy7g

### DIFF
--- a/configs/basic-single-node/k8s.yaml
+++ b/configs/basic-single-node/k8s.yaml
@@ -1,0 +1,7 @@
+head_node:
+  required_resources:
+    CPU: 8
+    memory: 32Gi
+  resources:
+    CPU: 0
+auto_select_worker_config: true
--- a/configs/basic-single-node/k8s.yaml
+++ b/configs/basic-single-node/k8s.yaml
@@ -1,7 +1,12 @@
 head_node:
   required_resources:
-    CPU: 8
-    memory: 32Gi
+    CPU: 4
+    memory: 8Gi
   resources:
     CPU: 0
-auto_select_worker_config: true
+worker_nodes:
+- name: cpu_worker
+  required_resources:
+    CPU: 8
+    memory: 32Gi
+  max_nodes: 1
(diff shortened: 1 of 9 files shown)


configs/basic-single-node/k8s.yaml
```diff
@@ -0,0 +1,12 @@
+head_node:
```
elliot-barn, line 1: can we rename them to azure.yaml to match the convention of the other templates?

PR #941, opened 2026-08-08
fix(depsets): expand on base depset, batch 1 (serve-stable-diffusion-v2, fintech_fraud_risk)

elliot-barn:
## Why

`raydepsets` silently ignores `source_depset:` on `expand` operations — the expand path only reads the `depsets:` key (defaults to `[]`), and no validation rejects the wrong key. Every expand entry in `dependencies/template.depsets.yaml` used `source_depset:`, so every template lock was compiled from its `requirements.txt` alone, completely unconstrained by the base image depset.

This is **batch 1** of a series converting the entries in YAML order (3 per PR):
- `serve_stable_diffusion_depset` (templates/serve-stable-diffusion-v2)
- `fintech_fraud_risk_depset` (templates/fintech_fraud_risk)

## What

Switch `source_depset:` → `depsets:` on the two expand entries and recompile their locks, which now carry the base image pins (verified: numpy/pandas/pydantic in each template lock match the base `ray_depset` lock exactly, which in turn matches pip freezes from the real anyscale ray images).

## langchain-agent-ray-serve dropped from this batch

It was originally converted here, but is **held back** along with all other `ray_llm`-based templates (audio-dataset-curation-llm-judge, multi_agent_a2a, llm_batch_inference_vision, llm_finetuning): the fetched OSS rayllm deplock pins the numpy-2 stack (`numpy==2.2.6` / `pyarrow==23.0.1` / `opencv 4.13`), while the actual `anyscale/ray-llm:2.56.0` image ships the numpy-1 stack (numpy 1.26.4 / pyarrow 19.0.1 — verified in #887's CI package diff). These templates layer the full lock onto the stock image with `uv pip install --no-deps`, so expanding on that base would force-upgrade numpy on the image and break its numpy-1-compiled scipy ABI. It's also not fixable by relaxing pins: `vllm==0.22.0 → opencv-python-headless>=4.13 → numpy>=2` makes a numpy-1 lock containing vllm unresolvable. The ray_llm templates need an image-accurate base lock before they can be converted.

## Testing

```
bash update_deps.sh --name serve_stable_diffusion_depset_2.56.0_3.12_cu128  # OK
bash update_deps.sh --name fintech_fraud_risk_depset_2.56.0_3.11            # OK
```

## Notes


Aydin-ab (approved):
Validated: /test-template passed for both templates (template-test #572).

PR #943, opened 2026-08-10
fix(depsets): expand on base depset, batch 3 (creatives_diffusion_finetuning, ecommerce_multi_model_serving, biotech_boltz_screening)

elliot-barn:
## Why

`raydepsets` silently ignores `source_depset:` on `expand` operations — the expand path only reads the `depsets:` key (defaults to `[]`). Every expand entry in `dependencies/template.depsets.yaml` used `source_depset:`, so every template lock was compiled from its `requirements.txt` alone, unconstrained by the base image depset.

Batch **3** of the series (see #941 for full context, #942 for batch 2):
- `creatives_diffusion_finetuning_depset`
- `ecommerce_multi_model_serving_depset`
- `biotech_boltz_screening_depset`

## What

- Switch `source_depset:` → `depsets:` on the three expand entries and recompile their locks, which now carry the base image pins (spot-checked numpy/pandas/pydantic match the base depset lock exactly).
- biotech_boltz_screening expands on a new `relax` depset that strips `scipy`/`pyyaml`/`requests` from the ray base: boltz exact-pins those (`scipy==1.13.1`, `pyyaml==6.0.2`, `requests==2.32.3`) against the base's 1.11.4/6.0.3/2.32.5. The template's requirements.txt already pins scipy==1.13.1, and the previous (standalone) lock shipped the same versions, so this preserves the template's verified behavior.

## Testing

```
bash update_deps.sh --name creatives_diffusion_finetuning_depset_2.56.0_3.11   # OK
bash update_deps.sh --name ecommerce_multi_model_serving_depset_2.56.0_3.11    # OK
bash update_deps.sh --name biotech_boltz_screening_depset_2.56.0_3.11          # OK
```

## Notes

- Not duplicating existing work: no open PR touches these depset entries.
- AI assistance was used for this change (Claude Code); a human reviewed the diff and ran the compile commands above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #942, opened 2026-08-08
fix(depsets): expand on base depset, batch 2/12 (tune_pytorch_asha, biotech_protein_embeddings, ecommerce_batch_embeddings)

elliot-barn:
## Why

`raydepsets` silently ignores `source_depset:` on `expand` operations — the expand path only reads the `depsets:` key (defaults to `[]`). Every expand entry in `dependencies/template.depsets.yaml` used `source_depset:`, so every template lock was compiled from its `requirements.txt` alone, unconstrained by the base image depset.

Batch **2 of 12** (see #941 for batch 1 and full context):
- `tune_pytorch_asha_depset`
- `biotech_protein_embeddings_depset`
- `ecommerce_batch_embeddings_depset`

## What

Switch `source_depset:` → `depsets:` on the three expand entries and recompile their locks, which now carry the base image pins.

## Testing

```
bash update_deps.sh --name tune_pytorch_asha_depset_2.56.0_3.12_cu128        # OK
bash update_deps.sh --name biotech_protein_embeddings_depset_2.56.0_3.11     # OK
bash update_deps.sh --name ecommerce_batch_embeddings_depset_2.56.0_3.11     # OK
```

All three compile; spot-checked numpy/pandas/pydantic in each template lock match the base depset lock exactly.

## Notes

- Not duplicating existing work: no open PR touches these depset entries.
- AI assistance was used for this change (Claude Code); a human reviewed the diff and ran the compile commands above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #944, opened 2026-08-10
fix(depsets): expand on base depset, batch 4 (finetune-stable-diffusion, vhol-ray-data, deepspeed_finetune)

elliot-barn:
## Why

`raydepsets` silently ignores `source_depset:` on `expand` operations — the expand path only reads the `depsets:` key (defaults to `[]`). Every expand entry in `dependencies/template.depsets.yaml` used `source_depset:`, so every template lock was compiled from its `requirements.txt` alone, unconstrained by the base image depset.

Batch **4** of the series (see #941 for full context; #942, #943 for earlier batches):
- `finetune_stable_diffusion_depset`
- `vhol_ray_data_depset`
- `deepspeed_finetune_depset`

## What

- Switch `source_depset:` → `depsets:` on the three expand entries and recompile their locks, which now carry the base image pins (spot-checked numpy/pydantic/requests/pandas match the base depset lock).
- deepspeed_finetune expands on a new `relax` depset that strips `pandas` from the ray base: the template deliberately ships `pandas==1.5.3` (vs base 2.3.3), matching its previous lock. (datasets' `fsspec`/`dill` ranges are satisfied by the base pins, so nothing else needed relaxing.)

## Testing

```
bash update_deps.sh --name finetune_stable_diffusion_depset_2.56.0_3.11  # OK
bash update_deps.sh --name vhol_ray_data_depset_2.56.0_3.11              # OK
bash update_deps.sh --name deepspeed_finetune_depset_2.56.0_3.11         # OK
```

## Notes


Aydin-ab:
/test-template deepspeed_finetune finetune-stable-diffusion vhol-ray-data

Aydin-ab (approved):
Validated: /test-template passed for all 3 templates (template-test #571).

PR #947, opened 2026-08-10
fix(depsets): expand on base depset, batch 7 (xgboost-training-and-serving, ray_train_workloads, ray-tune-train-integration)

elliot-barn:
## Why

`raydepsets` silently ignores `source_depset:` on `expand` operations — the expand path only reads the `depsets:` key (defaults to `[]`). Every expand entry in `dependencies/template.depsets.yaml` used `source_depset:`, so every template lock was compiled from its `requirements.txt` alone, unconstrained by the base image depset.

Batch **7** of the series (see #941 for full context; #942–#946 for earlier batches):
- `xgboost_training_and_serving_depset`
- `ray_train_workloads_depset`
- `ray_tune_train_integration_depset`

## What

- Switch `source_depset:` → `depsets:` on the three expand entries and recompile their locks, which now carry the base image pins (spot-checked numpy/pandas/pydantic match the base depset lock exactly).
- xgboost-training-and-serving (`mlflow==2.19.0` requires `pyarrow<19`) and ray_train_workloads (pins `pyarrow==14.0.2`) deliberately ship a pyarrow below the base's 19.0.1 — the same versions their previous standalone locks shipped (18.1.0 / 14.0.2) — so both expand on a new shared `relax` depset that strips `pyarrow` from the ray base.

## Testing

```
bash update_deps.sh --name xgboost_training_and_serving_depset_2.56.0_3.12  # OK
bash update_deps.sh --name ray_train_workloads_depset_2.56.0_3.12           # OK
bash update_deps.sh --name ray_tune_train_integration_depset_2.56.0_3.12    # OK
```

## Notes


Aydin-ab:
/test-template ray_train_workloads ray-tune-train-integration xgboost-training-and-serving

Aydin-ab (approved):
Validated: /test-template passed for all 3 templates (template-test #573).

PR #950, opened 2026-08-10
fix(depsets): expand on base depset, batch 10 (ray-summit-core-masterclass, text-embeddings, e2e-timeseries-forecasting)

elliot-barn:
## Why

`raydepsets` silently ignores `source_depset:` on `expand` operations — the expand path only reads the `depsets:` key (defaults to `[]`). Every expand entry in `dependencies/template.depsets.yaml` used `source_depset:`, so every template lock was compiled from its `requirements.txt` alone, unconstrained by the base image depset.

Batch **10 (final)** of the series (see #941 for full context; #942–#949 for earlier batches):
- `ray_summit_core_masterclass_depset`
- `text_embeddings_depset`
- `e2e_timeseries_forecasting_depset`

## What

- Switch `source_depset:` → `depsets:` on the three expand entries and recompile their locks, which now carry the base image pins (spot-checked numpy/pydantic/click match the base depset lock exactly).
- text-embeddings expands on a new `relax` depset stripping `packaging` from the ray base: its `langchain==0.1.17` (via `langchain-core<0.2`) needs `packaging<24.0` vs base 24.2; the previous standalone lock shipped 23.2.
- e2e-timeseries-forecasting expands on a new `relax` depset stripping `pyyaml`/`pandas`/`aiohttp`/`fsspec`: the template pins the first three below the base pins, and its `s3fs==2023.5.0` requires `fsspec==2023.5.0` exactly (base 2023.12.1) — all matching its previous standalone lock.

## Testing

```
bash update_deps.sh --name ray_summit_core_masterclass_depset_2.56.0_3.12  # OK
bash update_deps.sh --name text_embeddings_depset_2.56.0_3.11              # OK
bash update_deps.sh --name e2e_timeseries_forecasting_depset_2.56.0_3.12   # OK
```

## Notes

- Not duplicating existing work: no open PR touches these depset entries.
- AI assistance was used for this change (Claude Code); a human reviewed the diff and ran the compile commands above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #954, opened 2026-08-12
fix(depsets): expand ecommerce_end_to_end on the base depset

svc-template-updater:
Completes the `source_depset:` → `depsets:` sweep from #941-#950 for `ecommerce_end_to_end`, whose lock was never anchored to the Ray base image (raydepsets ignores `source_depset:` on expand entries, so it silently resolved from requirements.txt alone).

Lock grows 55 → 151 pins as it inherits the base image world; 13 packages move to base-image versions. numpy/pyarrow/pandas/scipy are unchanged — the template already pinned them to the base image values.

## Testing
`update_deps.sh` regeneration is idempotent (check-depsets will pass). `/test-template ecommerce_end_to_end` to follow.

## Caveats
The five ray-llm templates (`langchain-agent-ray-serve`, `audio-dataset-curation-llm-judge`, `multi_agent_a2a`, `llm_batch_inference_vision`, `llm_finetuning`) are deliberately NOT included. Chaining them fails to resolve, and the conflicts are real: the repo rayllm base lock claims numpy 2.2.6 / pyarrow 23.0.1 / mcp 1.1.3, while the deployed `ray-llm:2.56.0-py312-cu130` image ships numpy 1.26.4 / pyarrow 19.0.1. Anchoring to that lock would force an ABI-breaking numpy upgrade. The rayllm base lock needs to match the image first — infra fix, not per-template.


Aydin-ab (approved):
Anchors the last plain-ray template to the base depset; ray-llm templates correctly deferred until the rayllm base lock matches the deployed image.

PR #955, opened 2026-08-12
fix(depsets): anchor ray-llm templates to an image freeze

svc-template-updater:
Completes the anchoring work from #941-#950 and #954 for the five ray-llm templates, which chaining cannot fix.

## Why a freeze and not the base lock

A lock is a model of the recipe; a freeze is a measurement of the image. They diverge:

| source | agrees with the real image | packages it can't see |
|---|---|---|
| `rayllm_*.lock` | 179/252 (71%) | 172 |
| image freeze | 425/425 | 0 |

Three causes. Ray compiles the LLM lock with **py3.13** constraints, so it says `numpy 2.2.6` where the image ships `1.26.4`. `deplocks/llm/*` is Ray's CI *test* dependency set, not an image manifest — no `rayllm_img_*` is published. And the image is assembled with `pip install --no-deps`, so it ships `opencv 4.13` with `numpy 1.26.4` and `vllm 0.22` with `protobuf 4.25.8` — combinations **no resolver can emit**, so no lock can describe it.

Chaining therefore fails outright, and forcing it through with `relax` entries produces a lock that installs numpy 2.2.6 over the image's 1.26.4 — the ABI break these templates already hit once.

## What this does instead

Seeds each lock with a freeze of the image (from `docs.anyscale.com/base-images`, the same data `/inspect`'s `image_inventory.py` reads). uv treats the `-o` file as resolution preferences, so image versions hold unless a requirement forces a move — the same result as `pip install` in a workspace on that image.

The seed is the freeze **plus** the previous lock's pins for packages the image does not ship. Without that second half every recompile re-floats them to newest: regenerating without it moved `xxhash` 3.8.1 → 4.0.0 in four locks with no input change.

## Testing

All five pass `/test-template`; regeneration is idempotent, so `check-depsets` holds. Divergence from the real image, before → after:

| template | before | after |
|---|---|---|
| llm_finetuning | 99 | 15 |
| multi_agent_a2a | 48 | 9 |
| langchain-agent-ray-serve | 33 | 6 |
| llm_batch_inference_vision | 27 | 2 |
| audio-dataset-curation-llm-judge | 2 | 0 |

What remains is deliberate pins (a2a-sdk needs `protobuf>=5.29.5`; llamafactory wants `torch 2.10.0`). This also closes unnoticed drift: `llm_finetuning` was shipping pyarrow 21.0.0, protobuf 6.33.6, scipy 1.17.1, cffi 2.0.0 and pillow 11.3.0 over an image carrying 19.0.1, 4.25.8, 1.11.4, 1.17.1 and 10.4.0, with nothing pinning them.

## What an adversarial review changed here

An earlier revision deleted nine pins from `requirements.txt` as redundant, on the theory that the freeze now holds those versions. It does — but only as a *preference*, which a transitive requirement can still overrule, and those specific pins exist because that exact downgrade broke `build_openai_app` on multi-node (#906, #907) and tripped the numpy 1.x/2.x ABI guard (#887). They are restored; a preference is not a substitute for an incident fix.

## Caveats

The freeze is committed by hand here; #956 automates the refresh. Preferences are soft, so a dependency that genuinely requires `numpy>=2` would still upgrade it — worth a narrow ABI check later.

Claude-Session: https://claude.ai/code/session_01WQASugoLWZWj2Lp5AYRDka


Aydin-ab:
/test-template langchain-agent-ray-serve audio-dataset-curation-llm-judge multi_agent_a2a

Aydin-ab:
/test-template langchain-agent-ray-serve audio-dataset-curation-llm-judge multi_agent_a2a

elliot-barn (approved):

PR #959, opened 2026-08-13
ci(deps): refresh image freezes on a Ray bump, retire the rayllm depset

Aydin-ab:
Stacked on #955 — review that one first.

Keeps the image freezes current on a Ray bump, and retires the rayllm base depset that #955 made dead.

## 1. Keep the freezes current

#955 commits per-image freezes by hand; a Ray bump would leave them pointing at the previous version's image. Adds one step to `ray-base-locks.yaml` beside the base-lock recompile — the existing PR step already runs `git add dependencies/`, so refreshed freezes ride along and are reviewed with the locks. `dependencies/images/tracked-images.txt` lists which variants to track, so adding one is a single line. Source is the package list Anyscale publishes at `docs.anyscale.com/base-images`.

## 2. Retire the rayllm base depset

Nothing consumes it since #955, and it is actively misleading: ray's `deplocks/llm/*` is a py3.13-constrained **test** dependency set, not a description of the shipped image — 71% agreement with a real freeze. Leaving it invites the next person to chain an LLM template to it and hit the numpy 1.x→2.x ABI break. It was also the fanout's completeness signal; that job passes to the image freezes, which answer it more directly — a freeze exists only once Anyscale published the image, which is exactly what the per-template agents need.

Removes: the `ray_llm_depset` entry, `fetch-ray-llm-depsets.sh`, and two orphaned `rayllm_*.lock` files (~11k lines).

## 3. Four defects an adversarial review found

- **The two completeness gates had drifted, killing the retry loop.** The probe and the fanout resolver each carried their own copy, kept in step by a comment asking humans to. After tightening the resolver to require a freeze per tracked image, the probe still accepted any one — so on a partial image publish it judged the version complete, logged "nothing to do", and stopped refreshing the freezes the resolver was waiting on. Only a manual dispatch could restart it. Now one definition in `depset_versions.py`, imported by both. The existing tests mocked this logic out entirely, which is how it drifted unseen; five real ones added, including an assertion that both callers resolve to the same function.
- **A truncated fetch was committed as a valid freeze.** The download redirected straight into the destination, so a mid-transfer failure or an error page left a short file that still parsed, and the gate only checked existence. Every lock seeded from it would then float free. Now written to a temp file and required to hold a plausible package count before taking the real name.
- **`check-depsets` ignored `dependencies/images/`,** so a PR editing a freeze or the tracked list passed green without recompiling anything.
- **The abandoned-branch cleanup read a failed `gh pr list` as "no PR open"** and would have deleted the branch of an open PR, closing it.

The base-locks PR now also runs the completeness gate and states in its body whether pushing `ray-fanout-<v>` will be accepted — the refresh skips an unpublished image and exits 0, so the gap previously surfaced only when someone pushed the tag.

## Testing

- `refresh-image-freezes.sh 2.56.0` reproduces the committed freeze byte-for-byte.
- **All 10 tracked images fetch cleanly at Ray 2.57.0** against the live index — the automation is proven against the real next version, not a hypothetical. The `-cpu` fallback is load-bearing for 2 of them.
- `latest-depset-version.py`: resolves `2.56.0`; `--require 2.57.0` fails closed and names every missing freeze.
- 18 `prepare-base-locks` unit tests pass (13 existing + 5 new).
- `ray-bump-fanout.yaml` and `trigger-cursor-bump.py` are byte-identical to `main` across the whole stack — verified by diff, not assumption.

## Reviewer note

`complete_versions()` is the piece worth a second pair of eyes. It fails closed in every case I could construct; the residual risk runs the other way — if Anyscale permanently drops a tracked variant, the gate blocks until someone edits `tracked-images.txt`.

Claude-Session: https://claude.ai/code/session_01WQASugoLWZWj2Lp5AYRDka

---
Replaces #956, which GitHub auto-closed when its base branch was deleted on merge of #955 (and refused to reopen after the branch was rebased onto `main`). Same content, same commits — carries over @elliot-barn's approval on #956, so it needs a fresh stamp.


svc-template-updater (approved):
Same content Elliot approved as #956 — rebased onto main after #955 squash-merged.

PR #975, opened 2026-08-15
perf(tests): cut the five slowest template CI runtimes

svc-template-updater:
Five templates were spending most of their CI budget on work that proves nothing. Four are scale knobs set only in `tests.sh`, so the shipped template is unchanged; the fifth was downloading a dataset from a throttled host.

## What changed

| Template | Before | Bottleneck |
|---|---|---|
| `distributing-pytorch` | 40.1 min | CIFAR-10 from the canonical host at ~80 kB/s, once per GPU node |
| `entity-recognition-with-llms` | 41.5 min | LoRA SFT for 625 steps (1000 samples x 5 epochs) |
| `object-detection-video-processing` | 25.1 min | 2 training epochs at ~5 min each |
| `biotech_protein_embeddings` | — | corpus size |
| `intro-ray-libraries` | — | Tune samples and epochs |

`distributing-pytorch` is the only one that touches template content: one line pointing `datasets.CIFAR10.url` at the S3 copy of the same archive. torchvision verifies `tgz_md5` either way, and the notebook already reads its Ray Data half from a Ray example bucket — this just makes both halves consistent. It also fixes a 35-minute stall for anyone running the template by hand.

Every knob defaults to the value the template shipped with, so opening a notebook gives the full-scale run.

## Testing

Build 620 passed on `biotech_protein_embeddings` and `intro-ray-libraries`. Build 621 covers the other three.


Aydin-ab (approved):

Aydin-ab:
/test-template distributing-pytorch entity-recognition-with-llms object-detection-video-processing

Aydin-ab:
/test-template entity-recognition-with-llms e2e-rag-deepdive unstructured_data_ingestion

PR #976, opened 2026-08-15
chore(templates): move 18 GPU templates to their max CUDA image

Aydin-ab:
18 GPU templates sat on an older CUDA tag than their Python version supports — 5 on cu121, 1 on cu124, 12 on cu128, against maxima of cu128 (py311) and cu129 (py312). Each template's `BUILD.yaml` image and its depset's seed-freeze path move together; the `lock-image` check keeps them in step.

## Why only one lock changes

The base images ship **no torch and no `nvidia-*`** — the CUDA tag selects the system toolkit, not Python packages. The freezes for `py312-cu128` and `py312-cu129` are identical apart from a header comment. So bumping the image alone changes nothing resolvable, and the locks stay put.

The PyTorch wheel index is a separate axis, and it moves for `tensor_parallel_dtensor` only. `download.pytorch.org/whl/cu128` publishes torch from **2.7.0** and `cu129` from **2.8.0**; the other twelve templates pin torch older than that, so pointing them at the new index makes uv fall back to PyPI and silently drop the CUDA-tagged wheel (`torch==2.7.0+cu128` → `torch==2.7.0`). Moving those requires bumping torch itself, which is a real behaviour change and deserves its own PR.

## Testing

`pre-commit` clean, including `lock-image` and the `BUILD.yaml` validator. `tensor_parallel_dtensor`'s lock is the only one that changed: torch and torchvision keep their versions and move `+cu128` → `+cu129`, with the nvidia runtime libs going 12.8.x → 12.9.x.

Templates still need a `/test-template` round — worth doing before the 2.56 republish, since this changes the image every one of them runs on.

## Follow-up

Twelve templates pin torch below their image's CUDA index. Worth deciding, separately, whether to bring them forward.

### DIFF
--- a/BUILD.yaml
+++ b/BUILD.yaml
@@ -147,7 +147,7 @@
   owner_team: ray-serve
   dir: templates/model-multiplexing
   cluster_env:
-    image_uri: anyscale/ray:2.56.0-slim-py312-cu128
+    image_uri: anyscale/ray:2.56.0-slim-py312-cu129
   compute_config:
     GCP: configs/basic-single-node/gce.yaml
     AWS: configs/basic-single-node/aws.yaml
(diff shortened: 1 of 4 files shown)


Aydin-ab:
/test-template tensor_parallel_dtensor serve-stable-diffusion-v2 image-search-and-classification

Aydin-ab:
Reopening under the service account so it can be reviewed and merged.

PR #739, opened 2026-06-01
deps: depsets for special templates (summit, text-embeddings, timeseries)

elliot-barn:
Compiles hash-pinned `python_depset.lock` files and switches runtime `pip install` to the lock; papermill via `uv` in tests. Testing via CI `/test-template`.

| Template | torch in lock |
|---|---|
| e2e-timeseries-forecasting | torch==2.3.0 |
| ray-summit-core-masterclass | torch==2.7.0+cu128 |
| text-embeddings | torch==2.4.1+cu124 |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #977, opened 2026-08-15
ci(hooks): require == pins, and refresh them on every Ray bump

Aydin-ab:
Across 40 templates, **30% of requirement lines are not `==`** — 49 inequalities and 27 bare package names. Those locks re-resolve to whatever is newest whenever they're rebuilt, so a template's behaviour changes with nobody editing it, and CI passes because it tests the drift.

`==` alone isn't enough either. `text-embeddings` has shipped `torch<2.5`, `langchain==0.1.17` and `transformers==4.40.2` since the commit that created the file (#739, June). A pin set once and never revisited is as stale as no pin, just deterministically so.

## What changed

**`pin-style`**, a fourth check in the existing pre-commit hook, fails any requirement that is neither `==` nor carries a trailing comment saying why it can't be. The 99 that predate the rule are grandfathered in `dependencies/loose-pins-allowlist.txt` — a backlog, not config. The check reports how many remain and names entries it no longer needs, so the number only goes down. Green today; stops new drift immediately.

**The Ray bump wave now re-pins.** `bump-ray-version.md` gains a "Refresh the pins" step between the image bump and the lock recompile — that's the one recurring moment the whole fleet's dependencies get looked at, and the image has just changed underneath them.

For `torch` the ceiling is the image's CUDA wheel index, not PyPI: `cu128` publishes from 2.7.0 and `cu129` from 2.8.0. Pin below the floor and uv falls back to PyPI, the wheel silently loses its CUDA tag (`torch==2.7.0+cu128` → `torch==2.7.0`), and the lock still compiles.

## Testing

`pre-commit` clean whole-repo. Verified `pin-style` exits 1 on a newly added loose pin, 0 once it carries a justifying comment, and 0 on the repo as-is. Also fixed a stale claim in `bump-ray-version.md` that nothing checks the lock/image pairing — `lock-image` has since #968.

## Caveat

This lands the rule and the backlog. Actually pinning those 99 is separate work needing a `/test-template` round per wave, since each is a potential behaviour change.

### DIFF
--- a/.claude/skills/template/references/dependencies.md
+++ b/.claude/skills/template/references/dependencies.md
@@ -199,9 +199,36 @@ bundle instead of pinning the old image's freeze.
    workers/replicas keep running stale deps.
 4. Scan the lock diff for a framework package moving below its image version — the "Runtime skew" trap above.
 
+## How to pin
+
+**Every requirement is `==`, at the newest version that works on that template's image.**
+
+`>=` and bare names are not pins. They tell the resolver "anything newer is fine", so the lock
+silently re-resolves to whatever is current every time it is rebuilt — the template's behaviour
+changes with nobody editing it, and CI passes because it tests the drift. `<` is worse: an upper
+bound with no lower one freezes a template in the past. `text-embeddings` carried `torch<2.5`,
+`langchain==0.1.17` and `transformers==4.40.2` for a year that way.
+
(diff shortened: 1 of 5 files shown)


Aydin-ab:
Reopening under the service account so it can be reviewed and merged.

